PR #640, opened 2015-01-19
allow passing Authenticator instance to ThreadedAuthenticator

cochiseruhulessin:


minrk:
Can you provide an example?

PR #645, opened 2015-02-15
Run tox tests with python3.4

ticosax:


minrk:
thanks. I don't use tox, so I'll probably remone the tox.ini.

ticosax:
tox is actually quite useful to test things locally.
At the moment I'm trying to merge .travis.yml and tox.ini to remove the burden of double maintenance. If the outcome is ok, I'll make a PR.

minrk:
I hope you aren't planning to make a PR to cause travis to use tox. That would not be desirable.

I think I will still probably remove tox.ini. I don't maintain it, and never will because I don't use it.

ticosax:
I'm actually trying to spot a bug that I can reproduce only with py34. It would be a shame to debug it without tox. I think, it is a useful tool.
But that's your call.

PR #647, opened 2015-02-16
test memoryview shape info

minrk:


jasongrout:
I don't think this checks that you aren't making a copy.  Do you want to also assert something like `view2.obj is frame.data` (or however you get to the raw data object behind the frame).

PR #654, opened 2015-03-06
Fixed byte size computation for numpy array

splendido:
* `len(A)` returns the number of rows (which is 1024 in this case) while the total number of elements is to be retrieved with `A.size`
* the exact byte size of each element can be obtained with `A.itemsize` regardless of the actual `dtype` used for the array

with these changes the original size is correctly computed as 8388608 bytes which is exactly 8Mb


minrk:
Thanks

PR #656, opened 2015-03-07
Adds support for building libsodium on Windows.

AndreLouisCaron:
Libsodium compiles fine on Windows, even with VS2008 (for Python 2.7).  However, it requires a few tricks:
- VS2008 doesn't provide `<stdint.h>`, so we need to mock it (like libzmq already does);
- VS2008 doesn't support C99 `static inline` functions, but removing the `inline` when compiling it for PyZMQ seems to work fine;
- before version 1.0.1, libsodium uses the same `DLL_EXPORT` switch as libzmq, which prevented from compiling libsodium as a dependency of libzmq (since version 1.0.1, libsodium changed this macro to `SODIUM_DLL_EXPORT` and no longer conflicts with PyZMQ).

### DIFF
--- a/patch/stdint.h
+++ b/patch/stdint.h
@@ -1,0 +1,11 @@
+#ifndef _stdint_h__
+#define _stdint_h__
+
+#include "../bundled/zeromq/src/stdint.hpp"
+
+#define UINT8_MAX 0xff
+#define UINT16_MAX 0xffff
+#define UINT32_MAX 0xffffffff
+#define UINT64_MAX 0xffffffffffffffffull
+
+#endif /* _stdint_h__ */
--- a/setup.py
+++ b/setup.py
@@ -304,8 +304,8 @@ class Configure(build_ext):
             pjoin('backend', 'cython'),
             'devices',
         )]
-        if sys.platform.startswith('win'):
-            settings['include_dirs'].insert(0, 'patch')
+        if sys.platform.startswith('win') and sys.version_info < (3, 3):
+            settings['include_dirs'].insert(0, pjoin('buildutils', 'include_win32'))
         
         for ext in self.distribution.ext_modules:
             if ext.name.startswith('zmq.lib'):
@@ -456,14 +456,15 @@ class Configure(build_ext):
         # Windows:
         # 1) tell libsodium to export its symbols;
         # 2) prevent libsodium from defining C99 `static inline` functions
-        #    which aren't parsed correctly by VS2008;
+        #    which aren't parsed correctly by VS2008 nor VS2010;
         # 3) provide an implementation of <stdint.h> which is not provided in
         #    VS2008's "standard" library;
         # 4) link against Microsoft's s crypto API.
         if sys.platform.startswith('win'):
             libsodium.define_macros.append(('SODIUM_DLL_EXPORT', 1))
             libsodium.define_macros.append(('inline', ''))
-            libsodium.include_dirs.append('patch')
+            if sys.version_info < (3, 3):
+                libsodium.include_dirs.append(pjoin('buildutils', 'include_win32'))
             libsodium.libraries.append('advapi32')
         # register the Extension
         self.distribution.ext_modules.insert(0, libsodium)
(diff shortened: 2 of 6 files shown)


AndreLouisCaron:
I've only tested this against Python 2.7.9 on Windows.  I assume this needs testing on other supported Python versions!

setup.py
```diff
@@ -304,6 +304,8 @@ def init_settings_from_config(self):
             pjoin('backend', 'cython'),
             'devices',
         )]
+        if sys.platform.startswith('win'):
+            settings['include_dirs'].insert(0, 'patch')
```
minrk: it seems like this should just affect libsodium, yes?

setup.py
```diff
@@ -510,7 +525,8 @@ def bundle_libzmq_extension(self):
 
             # And things like sockets come from libraries that must be named.
 
-            libzmq.libraries.extend(['rpcrt4', 'ws2_32', 'advapi32'])
+            libzmq.libraries.extend(['rpcrt4', 'ws2_32', 'advapi32', 'libsodium'])
+            libzmq.library_dirs.append('build/temp.win32-2.7/Release/buildutils')
```
minrk: This is definitely not going to work anywhere but 2.7.

patch/stdint.h
```diff
@@ -0,0 +1,11 @@
+#ifndef _stdint_h__
```
minrk: This should be in `buildutils/include_win32`

minrk:
Thanks, this is super helpful. I think it's going to need a bit of help on Python 3, and I think some of the additions may not be necessary, but we can work it out.

AndreLouisCaron:
Cool!  I'll start by addressing the issues you raised and the. I'll try it on Python 3 and update it he pull request!

minrk:
Terrific, thanks.

setup.py
AndreLouisCaron: When I remove it, I get this error:

    monitoredqueue.c
    bundled\zeromq\include\zmq.h(78) : fatal error C1083: Cannot open include file: 'stdint.h': No such file or directory

The `stdint.hpp` it provides is used only internally.  I guess `libzmq` also depends on `<stdint.h>` being available to client applications...

setup.py
AndreLouisCaron: Fixed by copying similar code for `libzmq` in the top-level `bundled_settings` function.

patch/stdint.h
AndreLouisCaron: Fixed.

setup.py
minrk: >  I guess libzmq also depends on <stdint.h> being available to client applications...

That seems improbable, since it builds successfully without it in master.

AndreLouisCaron:
Updated the pull request.  Tested on Python 2.7.9 and 3.4.3.  I think I addressed all issues, let me know if there's anything else.

I'm not sure how to test all build variants (debug/release/64-bit).  Do you have a procedure somewhere so that I can try them out or are you going to test it yourself?

minrk:
You can look at the script in `tools/release_windows.bat`, which does the Windows binary builds. It could be a starting point for testing.

setup.py
```diff
@@ -499,6 +513,8 @@ def bundle_libzmq_extension(self):
             # include defines from zeromq msvc project:
             libzmq.define_macros.append(('FD_SETSIZE', 1024))
             libzmq.define_macros.append(('DLL_EXPORT', 1))
+            libzmq.define_macros.append(('_CRT_SECURE_NO_WARNINGS', 1))
+            libzmq.define_macros.append(('ZMQ_USE_SELECT', 1))
```
minrk: why have you changed the value for ZMQ_USE_SELECT?

minrk:
Thanks, this is looking good. I'll try to find some time to test it (I may not be able to for the next two weeks, due to conferences and travel).

setup.py
AndreLouisCaron: When I don't set this, I get the following error:

    c:\users\acaron\desktop\develop\pyzmq\bundled\zeromq\src\poller.hpp(42) : fatal error C1189: #error :  None of the ZMQ_USE_* macros defined

I get this when building from master with out my changes too.  The one thing I'm doing differently is building against the latest ZeroMQ master, but I can't understand why that's creating the error.  Any ideas?

minrk: > The one thing I'm doing differently is building against the latest ZeroMQ master, but I can't understand why that's creating the error.

That could easily be the reason. Perhaps this changed in libzmq master.

setup.py
AndreLouisCaron: I just checked and it seems `zmq.h` in 4.0.x and 4.1.x provide their own definitions of integer types but master does not.  See:
- [zmq.h line 73 in zeromq/libzmq](https://github.com/zeromq/libzmq/blob/master/include/zmq.h#L73) -- no integer types,
- [zmq.h line 73 in zeromq/zeromq4-x](https://github.com/zeromq/zeromq4-x/blob/master/include/zmq.h#L73) -- integer types; and
- [zmq.h line 73 in zeromq/zeromq4-1](https://github.com/zeromq/zeromq4-1/blob/master/include/zmq.h#L73) -- integer types.

I'm guessing master is "wrong" here, but the stable fork's definitions are kind of strange (is it even legal C to use preprocessor directives to see if a `typedef` has been defined?)

setup.py
AndreLouisCaron: I haven't had time to test this yet, but since `ZMQ_USE_SELECT` is the only poller choice available on Windows, I guess it doesn't hurt to provide this definition.

AndreLouisCaron:
Hi, did you have time to try this out?

Thanks!

minrk:
@AndreLouisCaron sorry, not yet. I'm in the middle of a big project on IPython that's got some time pressure. I'll try to get to this soon. Thanks for your patience.

minrk:
@AndreLouisCaron I was able to do a bit more review and testing, and it all looks and works great. Thanks!

PR #663, opened 2015-04-11
Add warnings about setsockopt delayed effect.

dwf:
Adds two warnings about the fact that socket options (in particular, HWM) only takes effect on subsequent binds/connects, in places I would have seen it while trying to diagnose a problem I was having with memory usage.

This is mentioned in the [zeromq documentation](http://api.zeromq.org/2-1%3azmq-setsockopt) but nowhere in the guide, and is the source of [much confusion](http://stackoverflow.com/questions/8092473/zeromq-high-water-mark-doesnt-work). I figure adding it here can't hurt to repeat it as I don't see this behaviour changing.

### DIFF
--- a/zmq/backend/cython/socket.pyx
+++ b/zmq/backend/cython/socket.pyx
@@ -298,6 +298,13 @@ cdef class Socket:
         
         optval : int or bytes
             The value of the option to set.
+
+        Notes
+        -----
+        .. warning::
+
+            All options other than zmq.SUBSCRIBE, zmq.UNSUBSCRIBE and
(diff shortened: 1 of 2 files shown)


dwf:
Build failures are unrelated.

minrk:
Sure. I try not to have too much general zmq usage in the pyzmq docs, which really belong in the upstream zmq docs, but this seems fine.

dwf:
Yep, in general I am against a proliferation of duplicate documentation, but for "gotchas" like this, practicality beats purity in my book.

minrk:
Absolutely.

PR #664, opened 2015-04-12
Use a better range to bind to a random port on Windows

ccordoba12:
This PR is motivated by spyder-ide/spyder#1880, where a user mentioned that the current range of ports is also used by Windows Vista/7/8 to create dynamics connections.

This new range avoids the current (and frequent) failures some users observe when starting IPython kernels on this OS.

### DIFF
--- a/zmq/sugar/socket.py
+++ b/zmq/sugar/socket.py
@@ -175,7 +175,7 @@ class Socket(SocketBase, AttributeSetter):
     
     getsockopt_unicode = getsockopt_string = get_string
     
-    def bind_to_random_port(self, addr, min_port=49152, max_port=65536, max_tries=100):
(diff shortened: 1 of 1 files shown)


minrk:
The current port range is the IANA recommended value. What's the source for the Windows values? What values are problematic on Windows?

ccordoba12:
> What's the source for the Windows values?

I really don't know. I just took our user's recommendation. I'll investigate further.

> What values are problematic on Windows?

The current ones, according to him. Truth is I've been unable to reproduce this bug but we receive complaints about it from time to time.

minrk:
I don't think this problem affects master, with the changes in #651 which asks the system to bind to a random port by using port 0, rather than trying random ports. I'll cut 14.6 with that fix soon.

ccordoba12:
Should I close this PR then?

minrk:
Yup, I think we can close this.

PR #677, opened 2015-06-05
update bundled libzmq to 4.1.2

minrk:
and include checksum of bundled dependencies for sanity, especially since libzmq comes over http.


minrk:
This is passing, but I don't want to merge it before 4.1.2 is out.

PR #688, opened 2015-06-08
test with multiple libzmq repos

minrk:


minrk:
Allows verification that we are working with the maintenance branches. This would have caught bugs backported to 4.0.6.

PR #691, opened 2015-06-10
Only consider debug version of bundled libraries if in debug interpreter

aburgm:
Without this, if there are both libzmq.pyd and libzmq_d.pyd around, one of
the two would be taken by chance. This patch makes sure that when running
in a debug interpreter we always load libzmq_d.pyd, and in a normal
interpreter we load libzmq.pyd.

It might make sense to instead keep two completely different python environments for debug and non-debug around, in which case this would not be needed. On the other hand, regular python import statements work with an environment where there are both debug and non-debug modules in the same tree, so maybe pyzmq should as well.

I tested this only on Windows.

### DIFF
--- a/zmq/__init__.py
+++ b/zmq/__init__.py
@@ -17,6 +17,23 @@ for ext in ('pyd', 'so', 'dll', 'dylib'):
     bundled_sodium.extend(glob.glob(os.path.join(here, 'libsodium*.%s*' % ext)))
     bundled.extend(glob.glob(os.path.join(here, 'libzmq*.%s*' % ext)))
 
+# If we are running in a debug interpreter, load libzmq_d.pyd instead of libzmq.pyd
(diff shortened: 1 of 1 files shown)


aburgm:
Also, in case you are running in a debug interpreter and don't have libzmq_d.pyd available but only libzmq.pyd, then, before, it tried to import libzmq.pyd and failed with a dialog box that MSVCR.DLL could not be opened. With this patch, the error message is simply

```
ImportError: DLL load failed: The specified module could not be found.
```

minrk:
Can you add a condition that only does this on Windows? It doesn't make sense on other platforms.

aburgm:
Okay, wasn't sure whether this is an issue on non-Windows as well. I updated the commit.

tomspur:
Why is this only an issue on windows?
There are also debug versions of python on Linux. There are just very few debug versions of modules packaged.

minrk:
The filename change is only relevant on Windows.

PR #692, opened 2015-06-12
retry interrupted system calls

### LINKED ISSUE #679: Retry system calls automatically
Now that Python finally retries system calls automatically (https://www.python.org/dev/peps/pep-0475/) it would be great if the zmq API would behave in the same way and sensibly re-tries interrupted system calls.

minrk:
cc @flub

- cython backend only so far
- retry on EINTR for send,recv,poll,device
- are there other calls that should be retried?

TODO:

- [x] tests
- [x] cffi backend
- [x] determine if there are other calls that might need to be retried (e.g. get/setopt, msg/ctx operations)

closes #679

### DIFF
--- a/zmq/backend/cython/checkrc.pxd
+++ b/zmq/backend/cython/checkrc.pxd
@@ -2,6 +2,10 @@ from libc.errno cimport EINTR, EAGAIN
 from cpython cimport PyErr_CheckSignals
 from libzmq cimport zmq_errno, ZMQ_ETERM
 
+cdef enum int:
+    SUCCESS = 0
+    RETRY_SYS_CALL = 1
+
 cdef inline int _check_rc(int rc) except -1:
     """internal utility for checking zmq return condition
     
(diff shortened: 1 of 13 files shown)


flub:
Thanks for getting to this so quickly!

Looking at the manpages getsocketop/setsockopt can raise EINTR as well, it looks like msg/ctx operations are all fine though.

zmq/backend/cython/checkrc.pxd
```diff
@@ -1,7 +1,12 @@
+from libc.stdio cimport printf
```
flub: I've never worked with cython before, but I don't think this is being used?  Probably a left over from some debugging you did?

minrk:
Okay, get/setsockopt are retried now as well, and there are tests for send/recv/poll. Just the CFFI implementation to go.

minrk:
Okay, I think this is good to go, now. Anything more that you can think of?

Issue #700, opened 2015-06-23
pkg-config used to get prefix for libzmq, but should be used to get full paths


fwph:
pkg-config is currently used in setup.py like:

    pkg-config --variable=prefix --print-errors libzmq

But pkg-config can be used to get all the information needed to build against libzmq, rather than building the cflags/ldflags from the prefix. Attempting to construct it by doing

    pjoin(prefix, 'lib')

Can just lead to sadness when building in an environment with non-standard library dirs (as I've been doing, on a system that places 64-bit libraries into $PREFIX/lib64). It would be better to just use *all* the information pkg-config can provide in this case. (I may be able to provide a patch, but I need to check on some employer policies first; it's also a fairly simple exercise to get it basically working)

minrk:
Makes sense. I've never encountered a case where this is needed, so it will probably be a low priority, as I will have no way to test if I've done it correctly.

fwph:
I got approval, so I'll provide a patch shortly.

minrk:
closed by #701

PR #701, opened 2015-06-25
updates setup.py to use pkgconfig to get the libdir and includedir instead of the prefix

fwph:
Update calls out to pkg-config in a slightly more systematic way to first check if zmq is present, then get the libdir, includedir, and library name (a little ugly; based off the flags to reduce the number of calls) instead of trying to derive this information from the prefix. Tested with OS 10.10, Ubuntu trusty, with both packaged (homebrew/apt) and locally built versions of zeromq, with and without pkg-config present.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -177,7 +177,7 @@ def check_pkgconfig():
                 zmq_config['include_dirs'].append(tok[2:])
             if tok.startswith("-l"):
                 zmq_config['libraries'].append(tok[2:])
-        info(zmq_config)
+        info("Settings obtained from pkg-config: %r" % zmq_config)
 
     return zmq_config
 
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ from unittest import TextTestRunner, TestLoader
 from glob import glob
 from os.path import splitext, basename, join as pjoin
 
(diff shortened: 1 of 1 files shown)


fwph:
(fix for issue #700 )

setup.py
```diff
@@ -148,6 +148,38 @@ def bundled_settings(debug):
...
+    except OSError as osexception:
+        if osexception.errno == errno.ENOENT:
+            info('pkg-config not found')
+        else:
+            warn("Running pkg-config failed - %s." % osexception)
+    except CalledProcessError:
+        info("Did not find libzmq via pkg-config.")
+
+    if pcfg is not None:
+        output, _ = pcfg.communicate()
```
minrk: probably want to add `output = output.decode('utf8', 'replace')` on this one, for py3 compat.

minrk:
Thanks, I appreciate anyone willing to wade through the mess I've made of setup.py. Looks like there's a small amount of bytes-unicode stuff needed for Python 3 to work.

fwph:
Oops, of course I forget python3. Hopefully that should fix it!

minrk:
Thanks!

PR #713, opened 2015-08-03
Fix undefined variable tout in zmq.green

### LINKED ISSUE #712: Undefined variable when polling with gevent.Timeout
When a gevent.Timeout() is used with zmq.green.Socket.poll() (with timeout of -1) and the timeout fires before the poll() timeout, the following exception occurs:

```
  File "polltest.py", line 10, in <module>
    sock.poll(timeout=None, flags=2)
  File "/usr/lib/python2.7/site-packages/zmq/sugar/socket.py", line 501, in poll
    evts = dict(p.poll(timeout))
  File "/usr/lib/python2.7/site-packages/zmq/green/poll.py", line 89, in poll
    if t is not tout:
UnboundLocalError: local variable 'tout' referenced before assignment
```

Here is the test program to reproduce the issue:

```python
import gevent
from zmq import green

sock = green.Context.instance().socket(green.DEALER)
sock.immediate = True
sock.connect('ipc://@/somewhere/over/the/rainbow')

timer = gevent.Timeout.start_new(0.01)
sock.poll(timeout=None, flags=2)
```

I looked at the code of the offending method and will submit a pull request shortly.

hashstat:
closes #712

### DIFF
--- a/zmq/green/poll.py
+++ b/zmq/green/poll.py
@@ -64,6 +64,8 @@ class _Poller(_original_Poller):
 
         if timeout > 0:
             tout = gevent.Timeout.start_new(timeout/1000.0)
+        else:
+            tout = None
 
         try:
             # Loop until timeout or events available


minrk:
Thanks

PR #730, opened 2015-10-10
don't provide <stdint.h> in non-bundle mode.

jizhilong:
though vs2008 has no support for `stdint.h`, providing
our own implementation will not work if we are not
building in bundle mode, since there will be no
"bundled/zeromq/stdint.hpp" exists, which is hardcoded in
buildutils/include_win32/stdint.h

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -337,8 +337,17 @@ class Configure(build_ext):
             pjoin('backend', 'cython'),
             'devices',
         )]
+        # provide an implementation of <stdint.h> if we are building in
+        # bundle mode on Windows.
         if sys.platform.startswith('win') and sys.version_info < (3, 3):
-            settings['include_dirs'].insert(0, pjoin('buildutils', 'include_win32'))
+            if cfg['libzmq_extension']:
+                settings['include_dirs'].insert(0,
+                                 pjoin('buildutils', 'include_win32'))
+            else:
+                warn("<stdint.h> is not provided by VS2008, which is a must"
+                     "for building pyzmq, you can find a solution on"
+            "http://stackoverflow.com/questions/126279/c99-stdint-h-header-and-ms-visual-studio"
+                    )
         
         for ext in self.distribution.ext_modules:
             if ext.name.startswith('zmq.lib'):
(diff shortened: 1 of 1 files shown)


minrk:
If I understand the warning correctly, it is suggesting that this does exactly the wrong thing on VS2008, which is what Python 2 will most often be using. Is that not correct?

jizhilong:
Hi, mirek, sorry for my poor English, which caused misunderstanding.I just updated the warning message, does it make sense now?

minrk:
Sorry, I still don't understand. As I read the code and the comment, it seems to mean:

1. it is choosing to not include stdint.h
2. it needs to include stdint.h to work on VS2008

Is that not correct?

jizhilong:
1. `stdint.h` is a C99 standard head file, but VS2008 has poor support for C99, and `stdint.h` is missing in VS2008 by default.
2. `stdint.h` is included in `zmq.h`, which is included by  `buildutils/vers.c`, so compiling `buildutils/vers.c' with msvc complaining `Can not find file stdint.h`.
3. To resolve this issue, a customized `stdint.h` was added in commit https://github.com/zeromq/pyzmq/commit/bdccfb646ba739da573781007ce86490512917e5, which include `stdint.hpp` from bundled zeromq.
4. However, if we build pyzmq in non-bundle mode, no zeromq source code will be downloaded, thus `#include ../../bundled/zeromq/src/stdint.hpp` will fail.So we should not try to provide our customized `stdint.h` in such case.

minrk:
When building a release of pyzmq, bundled libzmq source is always available, even if it's not used, so the file being unavailable doesn't work. If you are running from git, you can fetch it with `python setup.py fetch_libzmq`. If the only reason to skip it is that the file may not be available in an incomplete git checkout, that's not enough, especially if it will break builds with the official Python.org CPython 2.7, which uses VS2008.

jizhilong:
I encountered this issue while building pyzmq by runnning `python setup configure --zmq=../libzmq-develop` and `python setup.py bdist_msi`. Thanks for the tip `python setup.py fetch_libzmq`, it should resolve the issue.
I will close this PR, and since `bundled/zeromq` must exist no matter building on windows with bundled libzmq or not, I'd suggest to add a tip in https://github.com/zeromq/pyzmq/wiki/Building-and-Installing-PyZMQ

PR #737, opened 2015-10-28
Add Future-returning API for tornado

minrk:
instantiate `zmq.eventloop.future.Context` and blocking methods will return Futures.

to do:

- [x] Future for recv
- [x] Future for send (rarely useful)
- [x] Future for poll


zmq/eventloop/future.py
```diff
@@ -0,0 +1,309 @@
...
+        kwargs['flags'] |= _zmq.DONTWAIT
+        try:
+            result = recv(**kwargs)
+        except Exception as e:
+            f.set_exception(e)
+        else:
+            f.set_result(result)
+    
+    def _handle_send(self):
+        f = None
```
jizhilong: these lines can be simplified with the `while...else` syntax as below:
```
while self._send_futures:
    f, kind, kwargs, msg = self._send_futures.pop(0)
    if not f.done():
        break
else:
     return
```

zmq/eventloop/future.py
```diff
@@ -0,0 +1,309 @@
...
+                trigger_timeout
+            )
+            def cancel_timeout(f):
+                loop.remove_timeout(timeout_handle)
+            future.add_done_callback(cancel_timeout)
+        
+        def cancel_watcher(f):
+            if not watcher.done():
+                watcher.cancel()
+        future.add_done_callback(cancel_timeout)
```
jizhilong: is `cancel_timeout` a typo here? Did you actually mean `future.add_done_callback(cancel_watcher)`?

zmq/eventloop/future.py
```diff
@@ -0,0 +1,309 @@
...
+
+
+_FutureEvent = namedtuple('_FutureEvent', ('future', 'kind', 'args', 'msg'))
+
+
+class Poller(_zmq.Poller):
+    """Poller that returns a Future on poll, instead of blocking."""
+    def poll(self, timeout=-1):
+        """Return a Future for a poll event"""
+        future = Future()
```
jizhilong: this statement may cause `NameError` if `from concurrent.futures import Futurexx` succeed.

deltheil:
I have a question related to this new Future-returning API: at the time of writing, this new async socket registers into the I/O loop via [`io_loop.add_handler`](https://github.com/zeromq/pyzmq/blob/32bc51761a6fc0720540bafe248c0c3338d61ff4/zmq/eventloop/future.py#L309) but it never removes itself.

Is it normal?

On the application side (Tornado) I worked around this problem with an explicit call:

```Python
socket.io_loop.remove_handler(socket)
socket.close()
```

Shouldn't you override the `close` method so that it calls `io_loop.remove_handler` like [zmqstream does](https://github.com/zeromq/pyzmq/blob/32bc51761a6fc0720540bafe248c0c3338d61ff4/zmq/eventloop/zmqstream.py#L387-L391)?

PR #739, opened 2015-11-02
Fix Windows build for Python 2.7

cwacekINV:
- import setuptools in setup.py to take advantage of patches to
  distutils.find_vcvarsall

    Without this, `python setup.py install` fails due to the cryptic "Unable to find vcvarsall.bat". This isn't a problem with `pip install pyzmq` because pip monkeypatches setuptools in anyway.

- Revert bundled ZeroMQ version to 4.1.2 because of zeromq/libzmq#1630

   The packaged version of libzmq 4.1.3 contains an invalid version of the src/stdint.h file (one that doesn't actually exist in the git history anywhere). This is 100% a workaround, but its either that or patch the bundled libzmq, which seems more risky.

### DIFF
--- a/buildutils/bundle.py
+++ b/buildutils/bundle.py
@@ -33,10 +33,10 @@ pjoin = os.path.join
 # Constants
 #-----------------------------------------------------------------------------
 
-bundled_version = (4,1,3)
+bundled_version = (4,1,2)
 libzmq = "zeromq-%i.%i.%i.tar.gz" % (bundled_version)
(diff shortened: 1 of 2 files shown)


minrk:
It is not appropriate to import setuptools unconditionally, because it changes the behavior of `setup.py install` in unacceptable ways. If you want to build/install with setuptools, that should always be done with `pip install .`

minrk:
Merged as 270f0cc

PR #750, opened 2015-11-12
Add fileno() method do Socket class

felipecruz:
I was trying to use ZMQStream with tornado and noticed that the object passed to the event loop is a zmq socket class without a `fileno()` method. With this little path I got everything working as expected. Without I get:
```
TypeError: argument must be an int, or have a fileno() method.
```

### DIFF
--- a/zmq/sugar/socket.py
+++ b/zmq/sugar/socket.py
@@ -127,7 +127,15 @@ class Socket(SocketBase, AttributeSetter):
         ):
             keys.extend(collection)
         return keys
-    
+
+    #-------------------------------------------------------------------------
+    # File object interface fileno()
+    #-------------------------------------------------------------------------
+
+    def fileno(self):
+        return self.getsockopt(zmq.FD)
+
+
     #-------------------------------------------------------------------------
     # Getting/Setting options
     #-------------------------------------------------------------------------
--- a/zmq/tests/test_socket.py
+++ b/zmq/tests/test_socket.py
@@ -434,6 +434,11 @@ class TestSocket(BaseZMQTestCase):
         rcvd = self.recv(b)
         self.assertEqual(rcvd, b'hi')
 
+    def test_socket_fileno(self):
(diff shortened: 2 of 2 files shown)


minrk:
The zmq FD attribute shouldn't be used with pollers, and the FD attribute of a socket will *not* behave as you would expect, since it's an edge-triggered, read-only FD, not a level-triggered read-write FD. If you are encountering this with tornado, it's because you haven't told tornado to use the ZMQPoller, which you must do for it to work properly:

```python
from zmq.eventloop import ioloop
ioloop.install()
```

felipecruz:
Hello @minrk 

I'm actually calling the install function. I made this little example. Can you check it? It always throws me an error. I'm using Python 3.5 on osx

```python
import zmq

from zmq.eventloop.zmqstream import ZMQStream
from zmq.eventloop import ioloop as zmq_ioloop

from tornado import ioloop

zmq_ioloop.install()
loop = ioloop.IOLoop.instance()

ctx = zmq.Context()
pub = ctx.socket(zmq.PUB)
pub.bind('tcp://127.0.0.1:5000')

socket = ctx.socket(zmq.SUB)
socket.connect('tcp://127.0.0.1:5000')
socket.setsockopt_string(zmq.SUBSCRIBE, u"")

def echo(s, msg):
    print(msg)

stream = ZMQStream(socket)
stream.on_recv_stream(echo)

def send():
    pub.send_string('msg')
    print('sent {}'.format('msg'))
    loop.call_later(5, send)

loop.call_later(1, send)
loop.start()
```

minrk:
What version of pyzmq and/or tornado? That example is working for me right now with current stable of everything (Python 3.5, tornado 4.3, pyzmq 15).

felipecruz:
Hello @minrk 

I found the problem: I was calling `zmq_ioloop.install()` after `loop = ioloop.IOLoop.instance()`

Sorry to take your time..

PR #776, opened 2015-12-27
asyncio: fix event mask conversion with multiple events

### LINKED ISSUE #775: asyncio: aiohttp.request causes error after installing event loop
I have an app that uses aiohttp that I wanted to use pyzmq's new asyncio support with. However, it looks like the implementation causes errors when I try to use aiohttp's `request` method with an https address. Here's a simple test case:

```py
import asyncio
import aiohttp

import zmq
import zmq.asyncio


zmq.asyncio.install()

ctx = zmq.asyncio.Context()
sock = ctx.socket(zmq.PUB)
sock.bind('ipc:///tmp/something')


@asyncio.coroutine
def run():
    res = yield from aiohttp.request('GET', 'https://google.com')
    yield from res.text()

    yield from sock.send(b'received a thing')


if __name__ == '__main__':

    loop.run_until_complete(run())
```

This will result in the following traceback on Python 3.5.1:

```
  File "app.py", line 26, in <module>
    loop.run_until_complete(run())
  File "/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/base_events.py", line 325, in run_until_complete
    self.run_forever()
  File "/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/base_events.py", line 295, in run_forever
    self._run_once()
  File "/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/base_events.py", line 1218, in _run_once
    event_list = self._selector.select(timeout)
  File "/Users/tboyt/Code/asynczmq/venv/lib/python3.5/site-packages/zmq/asyncio.py", line 123, in select
    return [ (self.get_key(fd), _zmq2aio[evt]) for fd, evt in events ]
  File "/Users/tboyt/Code/asynczmq/venv/lib/python3.5/site-packages/zmq/asyncio.py", line 123, in <listcomp>
    return [ (self.get_key(fd), _zmq2aio[evt]) for fd, evt in events ]
KeyError: 5
```

It seems to me that the asyncio support assumes that the only async I/O you do is through ZMQ, and isn't set up to handle non-ZMQ sources of events. Is this an intended limitation?

One curious aspect of this issue is that a non-https request (e.g. `http://google.com`) does not result in this error. I suspect this has something to do with a difference in the underlying transport ([_SSLProtocolTransport](https://github.com/python/cpython/blob/master/Lib/asyncio/sslproto.py#L293) vs [_SelectorSocketTransport](https://github.com/python/cpython/blob/f0a4074ed747b21e9f36eab64e2467c359db6634/Lib/asyncio/selector_events.py#L624)).

minrk:
Fix KeyError when multiple events are available on one socket

closes #775


thomasboyt:
Oddly enough, this fix works for me on Python 3.5.1, but I still get a KeyError on Python 3.4.3. Not sure what the difference between the versions is.

minrk:
Can you show the full traceback you get from Python 3.4?

minrk:
Fixed the error on 3.4

PR #781, opened 2016-01-06
Asyncio ZMQSelector improvements and zap authentication

asteven:
The existing implementation of ZMQSelector was randomly failing on me with KeyErrors for unkown/unregistered file objects. The other existing asyncio.selectors ignore this error so I decided to rewrite ZMQSelector to behave alike.

This PR also adds a zap authenticator based on a asyncio task.

Consider this an RFC, I'm open for sugestions/ideas/changes.


zmq/asyncio.py
```diff
@@ -49,23 +49,35 @@ class _AsyncIO(object):
...
     _WRITE = selectors.EVENT_WRITE
     _READ = selectors.EVENT_READ
-    
+
     def _default_loop(self):
         return asyncio.get_event_loop()
 
 
-class ZMQSelector(selectors.BaseSelector):
+class ZMQSelector(selectors._BaseSelectorImpl):
```
minrk: It doesn't seem right to inherit from a private class, since we shouldn't assume it will continue to exist, even from one patch release to the next. Can this be done with public APIs?

minrk:
I'm not comfortable with using an undocumented private API as the base class, but otherwise this does seem like an improvement. A little work will need to be added to get the tests running, mostly adding the right skips for Python 2, etc.

zmq/asyncio.py
asteven: Agree. I went for the private class because it does all that is needed so was the simplest/fastest way to get a first version working.
No public APIs, would have to copy/reimplement code from selectors.py. No big deal, the whole implementation of _BaseSelectorImpl is ~ 90 lines of code including comments.

asteven:
Ok, I'll base on BaseSelector and implement the missing bits based on _BaseSelectorImpl.
Do you want the authenticator with examples and tests in a separate pull request or is it ok as is?

asteven:
I copied the missing bit's from python 3.5 selectors.py.
Do you know if this is allowed, and if yes if and how to properly attribute for license and copyright?

minrk:
The authenticator is fine. For copying code from stdlib, that's fine, too. Just add a comment at the top:

    # Derived from Python 3.5.1 selectors._BaseSelectorImpl, used under PSF License

minrk:
Thanks!

PR #803, opened 2016-02-04
Introduce decorators

iblis17:
Hi, we already has the `zmq.Context` and `zmq.Socket` act as a context manager, but there is deep indentation when we use both. So we can use decorator for function to make the code beautiful.
 - `context`
 - `socket`

For example.
```python
import zmq
from zmq.decorators import context, socket

@context()
@socket(zmq.PUB)
def f(skt, ctx):
    skt.bind('...')
    ...
```

For more usage, there are test cases in this PR.

Also, i'm willing to change variables naming, coding style, or making more documentation, if that make everyone comfortable to merge this PR.  :smile:

### DIFF
--- a/zmq/decorators/base.py
+++ b/zmq/decorators/base.py
@@ -50,24 +50,27 @@ class ZDecoratorBase(object):
 
                 self.hook('preinit')
 
-                with self.target(*self.dec_args, **self.dec_kwargs) as obj:
-                    self.hook('postinit')
+                try:
+                    with self.target(*self.dec_args, **self.dec_kwargs) as obj:
+                        self.hook('postinit')
 
-                    if self.kwname and self.kwname not in self.wrap_kwargs:
-                        self.wrap_kwargs[self.kwname] = obj
-                    elif self.kwname and self.kwname in self.wrap_kwargs:
-                        raise TypeError(
-                            "{0}() got multiple values for"
-                            " argument '{1}'".format(
-                                func.__name__, self.kwname))
-                    else:
-                        self.wrap_args += (obj,)
+                        if self.kwname and self.kwname not in self.wrap_kwargs:
+                            self.wrap_kwargs[self.kwname] = obj
+                        elif self.kwname and self.kwname in self.wrap_kwargs:
+                            raise TypeError(
+                                "{0}() got multiple values for"
+                                " argument '{1}'".format(
+                                    func.__name__, self.kwname))
+                        else:
+                            self.wrap_args += (obj,)
 
-                    self.hook('preexec')
-                    func(*self.wrap_args, **self.wrap_kwargs)
-                    self.hook('postexec')
-
-                self.hook('cleanup')
+                        self.hook('preexec')
+                        func(*self.wrap_args, **self.wrap_kwargs)
+                        self.hook('postexec')
+                except:
+                    raise  # re-raise the exception
+                finally:
+                    self.hook('cleanup')
 
             return wrapper
 
--- a/zmq/decorators/base.py
+++ b/zmq/decorators/base.py
@@ -1,0 +1,62 @@
+from functools import wraps
+
+from zmq.utils.strtypes import basestring
+
+
+class ZDecoratorBase(object):
+
+    def __init__(self, target):
+        self.target = target
+
+    def __call__(self, *dec_args, **dec_kwargs):
+        self.kwname = None
+        self.dec_args = dec_args
+        self.dec_kwargs = dec_kwargs
+
+        self.pop_kwname()
+
+        def decorator(func):
+            @wraps(func)
+            def wrapper(*wrap_args, **wrap_kwargs):
+                self.wrap_args = wrap_args
+                self.wrap_kwargs = wrap_kwargs
+
+                self.hook('preinit')
+
+                with self.target(*self.dec_args, **self.dec_kwargs) as obj:
+                    self.hook('postinit')
+
+                    if self.kwname and self.kwname not in self.wrap_kwargs:
+                        self.wrap_kwargs[self.kwname] = obj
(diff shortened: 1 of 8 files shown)


zmq/decorators/sugar.py
```diff
@@ -0,0 +1,66 @@
...
+        First, if there is an keyword argument named ``context`` and it is a
+        ``zmq.Context`` instance , we will take it.
+
+        Second, we check all the ``wrap_args``, take the first ``zmq.Context``
+        instance.
+
+        :raises TypeError: if any context instance can not be found
+        '''
+        if self.context_name in self.wrap_kwargs:
+            ctx = self.wrap_kwargs[self.context_name]
```
minrk: The default Context of the socket should probably be `zmq.Context.instance()`, so that socket-decorated methods do not also need to be context-decorated.

zmq/decorators/base.py
```diff
@@ -0,0 +1,62 @@
...
+                            " argument '{1}'".format(
+                                func.__name__, self.kwname))
+                    else:
+                        self.wrap_args = (obj,) + self.wrap_args
+
+                    self.hook('preexec')
+                    func(*self.wrap_args, **self.wrap_kwargs)
+                    self.hook('postexec')
+
+                self.hook('cleanup')
```
minrk: definitely cleanup and probably postexec are going to need to be in a finally block to ensure they are called. Though I see that they are not used, so perhaps remove them.

minrk:
Can you give an example (and test) of what this looks like to use two or more sockets, and the socket decorator without the context decorator?

zmq/decorators/sugar.py
iblis17: :-o done in c2612da

iblis17:
Hi @minrk !
Test cases for multi sockets added in ebe337d

zmq/decorators/base.py
iblis17: I place hooks here, and let the child class can use it.
I add more doc strings for `ZDecoratorBase`, please check out 10c33b9.

iblis17:
BTW, any design for `zmq.proxy`? It seems this is more complex case to design...

iblis17:
err.. I have no idea why ci failed..

zmq/tests/test_decorators.py
```diff
@@ -0,0 +1,156 @@
...
+
+    assert foo.type == zmq.PUB
+    assert bar.type == zmq.SUB
+    assert baz.type == zmq.PUSH
+
+
+@context()
+@socket(zmq.PUSH)  # baz
+@socket(zmq.SUB)   # bar
+@socket(zmq.PUB)   # foo
```
minrk: This seems backwards. Shouldn't the args be `(ctx, push, sub, pub)`? Is it possible to do that?

iblis17: I make the decorator prepend the arg just like [mock](https://docs.python.org/3/library/unittest.mock.html#quick-guide) do. I'm not sure what's the pros and cons.

And it's possible to make it in forward, and them will be placed between `args` and `kwargs`. e.g.:
```
@socket(zmq.PUB)
@socket(zmq.SUB)
def f(other_arg, pub, sub, other_kwargs=None):
    ...
```

Current implementation:
```
@socket(zmq.PUB)
@socket(zmq.SUB)
def f(sub, pub, other_arg, other_kwargs=None)"
    ...
```

Or... any suggestion?

iblis17: Hmm.. I will try to make `f(pub, sub, other_arg, other_kwars=None)` possible. ~~Maybe we just need some black magic~~

iblis17:
More tries about `mock.patch`.
I figure out that `mock.patch` does `f(otherargs, [mocked_args in backward ...], kwargs)`
```python
In [71]: @patch('os.path.isdir')
   ....: @patch('os.path.join')
   ....: def f(mock, join, isdir, fake=None):
   ....:     print(mock)
   ....:     print(join)
   ....:     print(isdir)
   ....:     print(fake)

In [72]: f('foo')
foo
<MagicMock name='join' id='34545165088'>
<MagicMock name='isdir' id='34543901272'>
None

In [73]: f('foo', 'bar')
foo
bar
<MagicMock name='join' id='34543967480'>
<MagicMock name='isdir' id='34543964960'>

In [74]: f('foo', fake='bar')
foo
<MagicMock name='join' id='34556771576'>
<MagicMock name='isdir' id='34556771744'>
bar
```

Then.

I tried to make the pattern `f([forward_args...], other_args, kwargs)`.
The Implementation became mush complex; it also brakes class method to pass `self` and let `functools.partial` fail.

So finally, i chose the pattern -- `f(other_args, [forward_args...], kwargs)`.

Please try the latest patch out. :)

minrk:
If you would prefer to follow mock's pattern exactly, that would be fine with me, since there's a popular library to point to for precedence.

There appear to be intermittent failures due to something funky in the threaded tests.

iblis17:
Oh. Before you highlighted the precedence, i never notice that injecting args in backward (like `mock` does) is so non-intuitive for most user. Maybe i should put alway my 'preconception'.

And the latest implementation is forward, it's intuitive to use, so i will vote for forward version.

which one do your vote for?
- forward: `f(other_args, [args in forward...], kwargs)`
- backward: (`mock.patch`-like): `f(other_args, [args in backward...], kwargs)`

----
About the threaded testing:
When i wrote the feature code, i do not take care the thread safety. And i hope the related problem solved in  6d18c4f. Please review my code, and i can explain them if you need.

minrk:
Forward makes more sense to me. Thanks!

iblis17:
:tada: I will write some documentation and send them later

PR #807, opened 2016-02-09
Simplify context/socket decorators

minrk:
- remove state from decorator instances (removes thread safety problems)
- single module, not pkg
- remove unused hooks


zmq/decorators.py
```diff
@@ -0,0 +1,187 @@
...
+                ...
+
+        '''
+        kw_name, dec_args, dec_kwargs = self.process_decorator_args(*dec_args, **dec_kwargs)
+
+        def decorator(func):
+            @wraps(func)
+            def wrapper(*args, **kwargs):
+                target = self.get_target(*args, **kwargs)
+                print(target)
```
iblis17: oops!

PR #814, opened 2016-02-29
Pass encoding and log parameters up to parent class

ballan91:
Encoding and log parameters are unused in the subclass, I think it's reasonable to pass them upstream.


minrk:
Thanks!

PR #816, opened 2016-03-03
Handle string input to z85

### LINKED ISSUE #815: zmq.utils z85 decode (of string) does not work with python 3
I have been using the z85 encoding to store hashes in config files and sqlite database columns, while using python 2 I was able to assume that a z85 encoded value was simply a string. In migrating to python 3 I found that I needed to specify z85 as bytes.

Is this inconsistent with other encoding libraries? For example:
```
>>> import hashlib, base64
>>> type(hashlib.sha1(b'wibble').hexdigest())
<class 'str'>
>>> type(hashlib.sha1(b'wibble').digest())
<class 'bytes'>
>>> type(base64.b64encode(hashlib.sha1(b'wibble').digest()))
<class 'bytes'> # WAT?
```

I'm not sure which one is the right way to do it.

```
>>> print("libzmq-%s" % zmq.zmq_version())
libzmq-4.1.2
>>> print("pyzmq-%s" % zmq.pyzmq_version())
pyzmq-15.2.0
>>> print("Python-%s" % sys.version)
Python-3.4.4 (v3.4.4:737efcadf5a6, Dec 20 2015, 19:28:18) [MSC v.1600 32 bit (Intel)]
```

Steps to reproduce:
```
>>> z85.decode("Z9]q$")
  File "<stdin>", line 1, in <module>
  File "F:\Python\python-3.4.4\lib\site-packages\zmq\utils\z85.py", line 55, in decode
    value += Z85MAP[z85bytes[i+j]] * offset
KeyError: 'Z'
```
Expected behaviour from python 2.7:
```
>>> z85.decode("Z9]q$")
'\xbe(\x96\xd5'
```

Now I know I can fix this by prepending b"" to my string and it works as expected but should the decode function check whether I am passing in bytes or a string and raise a specific Exception? Or should it just automatically convert a string to a bytearray?

Sekenre:
This fixes any confusion caused by passing a string in python 3 to the z85.decode function. Since python 3 treats strings differently from bytearrays, particularly in regards to doing list-type operations on them.

closes #815

### DIFF
--- a/zmq/utils/z85.py
+++ b/zmq/utils/z85.py
@@ -42,9 +42,12 @@ def encode(rawbytes):
         return b''.join(encoded)
 
 def decode(z85bytes):
-    """decode Z85 bytes to raw bytes"""
+    """decode Z85 bytes to raw bytes, accepts ASCII string"""
     if PY3 and isinstance(z85bytes, str):
-        raise TypeError("Unicode-objects must be encoded to bytes")
+        try:
+            z85bytes = z85bytes.encode('ascii')
+        except UnicodeEncodeError:
+            raise ValueError('string argument should contain only ASCII characters')
 
     if len(z85bytes) % 5:
         raise ValueError("Z85 length must be multiple of 5, not %i" % len(z85bytes))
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

PR #829, opened 2016-03-23
Include '*.pxi' files from Cython backend.

alexrudy:
When trying to write Cython code against pyzmq, I noticed that the *.pxi files don't get included in the distribution. It would be helpful to have them so that e.g. libzmq.pxd is a functional cython header.

This change just adds these files to package_data similar to how '*.pxd' files are included.


minrk:
Indeed! Goes to show how much this gets used.

PR #836, opened 2016-04-19
run tests with pytest

minrk:


minrk:
pytest's unconditional magic fixtures are super annoying here. I may drop pytest for nose2, instead.

iblis17:
Hi @minrk 
Just out of curiosity, what is the main perspective let you migrate to `pytest` ?

minrk:
@iblis17 I'm not quite sure I understand the question, but more recent tools (pytest, nose2) let us use plain assert statements, such as:

```python
assert a == b
```

instead of requiring special methods:

```python
self.assertEqual(a, b)
```

among a few other minor niceties. In general, it doesn't make much difference.

iblis17:
Yep. The assertion statements is more pythonic!
Is there any killer feature in `pytest` and let you drop `nose`?

minrk:
> Is there any killer feature in pytest and let you drop nose?

Nope, I just like it a bit more. It doesn't make a big difference.

iblis17:
Ha, OK! :smile:

setup.py
```diff
@@ -778,14 +758,9 @@ def run(self):
...
-        try:
-            import nose
-        except ImportError:
-            warn("nose unavailable, falling back on unittest. Skipped tests will appear as ERRORs.")
-            return self.run_unittest()
-        else:
-            return self.run_nose()
+        p = Popen([sys.executable, '-m', 'pytest', '-v', os.path.join('zmq', 'tests')])
+        p.wait()
+        sys.exit(p.returncode)
```
iblis17: OMG, how dusty this file is .... lots of trailing space here.

----

Hope this patch help you
See Also: https://pytest.org/latest/goodpractices.html#manual-integration
```diff
diff --git a/setup.py b/setup.py
index 3bed865..ca59aa6 100755
--- a/setup.py
+++ b/setup.py
@@ -756,52 +756,60 @@ class TestCommand(Command):
             "You must build pyzmq with 'python setup.py build_ext --inplace' for 'python setup.py test' to work.",
             "If you did build pyzmq in-place, then this is a real error."]))
             sys.exit(1)
-        
+
+        try:
+            import pytest
+        except ImportError:
+            print_exc()
+            fatal('Testing requirement "pytest" missing.')
+            sys.exit(1)
+
         info("Testing pyzmq-%s with libzmq-%s" % (zmq.pyzmq_version(), zmq.zmq_version()))
-        p = Popen([sys.executable, '-m', 'pytest', '-v', os.path.join('zmq', 'tests')])
-        p.wait()
-        sys.exit(p.returncode)
+
+        pytest_args = ['-v', os.path.join('zmq', 'tests')]
+        errno = pytest.main(pytest_args)
+        sys.exit(errno)
```

iblis17: If we migrate this `setup.py` from `distutils` to `setuptools`, we can remove the `try-execpt` block in this patch and just add `test_requires` option to `setup` function.

PR #839, opened 2016-04-20
Various fixes for hanging auth tests

### LINKED ISSUE #838: auth tests deadlocking on non-x86 and zmq > 4.0.5
recently libzmq in debian got updated to 4.1.4 and the pyzmq started deadlocking on several architectures, see here for the logs:
https://buildd.debian.org/status/package.php?p=pyzmq&suite=unstable

it is also reproducable with git head pyzmq and git head zeromq-4.1
the tests that usually lock up are in zmq.tests.test_auth, test_plain_unconfigured_server and test_plain_bogus_credentials

Apparently the commit in zeromq which triggers the problem is 

https://github.com/zeromq/libzmq/commit/9a53f334d25c27416b84d94a2d40e74656135b50
Don't reconnect on protocol errors

reverting only it in the stable branches it fixes pyzmq testsuite.

Do you maybe have some advice if the issue is really in that commit or pyzmq itself?

minrk:
I understand how this should fix hangs in the threaded socket tests, but I don't understand how or where the IOLoop tests would have ever hung, so a sleep is the best I could come up with.

I cannot verify the fix, because the hangs do not appear to be reproducible on x86 systems.

closes #838

### DIFF
--- a/zmq/auth/base.py
+++ b/zmq/auth/base.py
@@ -49,6 +49,7 @@ class Authenticator(object):
         self.zap_socket = self.context.socket(zmq.REP)
         self.zap_socket.linger = 1
         self.zap_socket.bind("inproc://zeromq.zap.01")
+        self.log.debug("Starting")
 
     def stop(self):
         """Close the ZAP socket"""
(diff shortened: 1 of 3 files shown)


minrk:
@juliantaylor if you could test this branch and report back, that would be super helpful.

juliantaylor:
this branch appears to fix the deadlock I reported earlier most ports now work
there is still a problem on arm64 but I haven't checked yet what is happening exactly.
https://buildd.debian.org/status/package.php?p=pyzmq&suite=unstable

minrk:
Thanks, @juliantaylor!

PR #843, opened 2016-04-25
Context.socket forwards kwargs to socket __init__

### LINKED ISSUE #840: Add ability to pass kwargs for Socket.__init__ to Context.socket
Some of the socket classes' constructors take keyword arguments in addition to socket_type; however, the `Context.socket` method only takes the `socket_type` argument. It would be useful to be able to pass keyword arguments to the socket constructor through `Context.socket`.

For example, `zmq.asyncio.Socket.__init__` accepts an event loop instance via the `io_loop` argument. I'm using pyzmq with asyncio and I would like to be able to explicitly pass an event loop instance to the socket via `Context.socket`.

I believe the simplest way to do this would be to add `**kwargs` to the method `zmq.sugar.context.Context.socket`, and to then pass that to the socket constructor.

If that sounds reasonable to the maintainers, I would be happy to open a PR.

leifurhauks:
closes #840

### DIFF
--- a/zmq/sugar/context.py
+++ b/zmq/sugar/context.py
@@ -125,7 +125,7 @@ class Context(ContextBase, AttributeSetter):
     def _socket_class(self):
         return Socket
     
-    def socket(self, socket_type):
+    def socket(self, socket_type, **kwargs):
         """Create a Socket associated with this Context.
 
         Parameters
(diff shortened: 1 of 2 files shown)


minrk:
Thanks!

PR #850, opened 2016-06-03
bump bundled libzmq-4.1.5pre

minrk:
For testing, mostly. 4.1.5 release should be coming pretty soon.


minrk:
Looks like the macro definitions for curve support have changed.

PR #862, opened 2016-07-08
always load bundled libzmq with import

### LINKED ISSUE #861: PyZMQ does not setup activation context for pyd extension on Windows
This is a bit of a long story, and it's not quite clear who's bug this is, but please bear with me while figuring this out.

On Windows, python and it's dependencies are generally linked against a specific named version of the MSCV C runtime (mscr90.dll). Making this work properly in general when parts of the application are using different C runtime versions requires a Side-by-Side manifest file. Libpython itself is built with such a manifest file and further, but extensions are generally not. Now, when libpython itself loads a `.pyd` extension, it extends its own activation context to the loaded extension [1], so everything is fine. 

Now, on Windows, pyzmq loads the `libzmq.pyd` python extension, but goes through ctypes [2] rather than the python extension mechanism, so the automatic extension of the activation context does not apply.

We encounter this in the python/julia bridge because julia is generally linked against a different C library.

Possible solutions:
1. Build libzmq.pyd with an appropriate manifest
2. Teach ctypes how to load libraries with the extended activation context.
3. Make ctypes handle the extended activation context by default for any shared library call
4. Don't use ctypes directly for loading libzmq.pyd (or at least load it via the standard extension mechanism first).
5. Work around this by setting the activation context manually in pyzmq
6. Work around this in the the embedded (in our case the python/julia bridge)

I will likely do approach 6 for the python/julia bridge to allow the basic use of `import zmq` to work, but I don't think it's the right long term solution (esp, because we can't really know when the right time to set the activation context is, so it will definitely be too broad). If approach 4 works, I think it would be a good fix to have in pyzmq (I wasn't able to figure out how to import a .pyd file given a file path, but would be happy to test if somebody can tell me). Approach 1 seems workable as well, but I don't know how that extension is built, so I'm not sure what it would take. Approaches 2/3 are more for the python/ctypes developers to think about, since I don't have any insight into what that would take.

Some references https://github.com/stevengj/PyCall.jl/issues/87 https://github.com/JuliaLang/julia/pull/17136 https://bugs.python.org/issue17213

[1] https://github.com/python/cpython/blob/c1ddef67a68a23e468984413b951097f8e2b63f3/Python/dynload_win.c#L216
[2]  https://github.com/zeromq/pyzmq/blob/a960cc659c3e8deb83a9219059099de61d3e55f6/zmq/__init__.py#L37

minrk:
instead of ctypes

since bundled libzmq is built as an Extension, this should always work.

closes #861

### DIFF
--- a/zmq/__init__.py
+++ b/zmq/__init__.py
@@ -6,7 +6,7 @@
 # load bundled libzmq, if there is one:
 def _load_libzmq():
     """load bundled libzmq if there is one"""
-    import sys, ctypes
+    import sys, ctypes, platform
     dlopen = hasattr(sys, 'getdlopenflags') # unix-only
     if dlopen:
         dlflags = sys.getdlopenflags()
@@ -18,6 +18,10 @@ def _load_libzmq():
     else:
         # store libzmq as zmq._libzmq for backward-compat
         globals()['_libzmq'] = libzmq
+        if platform.python_implementation().lower() == 'pypy':
+            # pypy needs explicit CDLL load for some reason,
+            # otherwise symbols won't be globally available
+            ctypes.CDLL(libzmq.__file__, ctypes.RTLD_GLOBAL)
     finally:
         if dlopen:
             sys.setdlopenflags(dlflags)
--- a/zmq/__init__.py
+++ b/zmq/__init__.py
@@ -3,53 +3,11 @@
 # Copyright (C) PyZMQ Developers
 # Distributed under the terms of the Modified BSD License.
 
-import os
-import sys
-import glob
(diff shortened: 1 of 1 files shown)


minrk:
Well, apparently that was just wrong. Will investigate.

minrk:
Interestingly, this works on OS X and Windows, but apparently not on Linux. I think I've got a version that works everywhere, now.

minrk:
Works everywhere *except* pypy on Linux. Getting closer...

minrk:
Seem to have figured it out.

Issue #845, opened 2016-05-18
Program no longer exits when using pyzmq 15.2.0


damonlynch:
Please accept my advance apologies for the truly vague nature of this bug report!

The problem is this: my program works well when using < pyzmq 15.2.0. When using pyzmq 15.2.0, it often but not always fails to exit, necessitating the use of Ctrl-C. 

Of course it could well be a bug in my code that pyzmq 15.2.0 is exposing. Or it might be a problem in pyzmq 15.2.0. How do I go about diagnosing the problem? FWIW my code uses Python 3.4/3.5, PyQt 5.4/5.5, and ZeroMQ 4.1.2 on Linux.

Thank you.

minrk:
Can you tell what call is hanging on exit? It's hard to track down without more information. Typically a hang on exit is due to sockets left open.

damonlynch:
I'm not sure how to even diagnose properly what is left hanging, because when I hit Ctrl-C, there is no error message of any kind, let alone one about any socket being open. However running netstat returned something of possible interest, but I lack the experience to understand if there is anything truly useful (for instance I have no idea where the [ACC] STREAM LISTENING part comes from) :

```
$ netstat --all --program | grep '5988'
(Not all processes could be identified, non-owned process info
 will not be shown, you would have to be root to see it all.)
tcp        0      0 *:39461                 *:*                     LISTEN      5988/python3    
tcp        0      0 *:37477                 *:*                     LISTEN      5988/python3    
tcp        0      0 *:36455                 *:*                     LISTEN      5988/python3    
tcp        0      0 *:35433                 *:*                     LISTEN      5988/python3    
tcp        0      0 *:36202                 *:*                     LISTEN      5988/python3    
tcp        0      0 *:35530                 *:*                     LISTEN      5988/python3    
tcp        0      0 *:36843                 *:*                     LISTEN      5988/python3    
tcp        0      0 *:44208                 *:*                     LISTEN      5988/python3    
tcp        0      0 *:33520                 *:*                     LISTEN      5988/python3    
tcp        0      0 *:45554                 *:*                     LISTEN      5988/python3    
tcp        0      0 *:38163                 *:*                     LISTEN      5988/python3    
tcp        0      0 *:40021                 *:*                     LISTEN      5988/python3    
tcp        0      0 *:45464                 *:*                     LISTEN      5988/python3    
tcp        0      0 *:34810                 *:*                     LISTEN      5988/python3    
[...]

Issue #837, opened 2016-04-19
linger argument on context.destroy() appears to do nothing


orodbhen:
According to the documentation 

    If linger is specified, the LINGER sockopt of the sockets will be set prior to closing.

However, I found that I have to set linger explicitly with setsockopt, or `context.destroy()` will hang indefinitely:

`````
c = zmq.Context()
s = c.socket(zmq.REQ)
s.connect("tcp://127.0.0.1:50000")
s.setsockopt(zmq.LINGER, 0)
print "sending message"
s.send("hello")
print "message sent"
if s.poll(1000):
    print s.recv()
else:
    print "timed out"
print "closing socket"
s.close()
print "destroying context"
c.destroy(0)
print "done"
`````

PR #865, opened 2016-07-16
cffi: fix linger in Socket.close() and Context.destroy()

dovf:
This changeset fixes bugs regarding how 'linger' was handled in the cffi
backend. Prior to this fix, running the following example would exit
immediately (as expected) when run under the cython backend, but would hang
under the cffi backend (`PYZMQ_BACKEND=cffi`).

```
import zmq

ctx = zmq.Context()
s = ctx.socket(zmq.REQ)
s.connect('tcp://localhost:34567')
s.send('test')

s.close(linger=0)      # <- cffi backend hangs here...
ctx.destroy(linger=0)  # <- ... and here
```

Context.destroy() did not correctly handle the case where linger=0, in which
case linger evaluates to False. The fix is to check specifically for None, not
for any non-True value.

Socket.close() was completely ignoring the 'linger' argument.

### DIFF
--- a/zmq/backend/cffi/context.py
+++ b/zmq/backend/cffi/context.py
@@ -96,7 +96,7 @@ class Context(object):
         for s in sockets:
             s = s()
             if s and not s.closed:
-                if linger:
+                if linger is not None:
                     s.setsockopt(LINGER, linger)
                 s.close()
         
--- a/zmq/backend/cffi/socket.py
+++ b/zmq/backend/cffi/socket.py
@@ -90,6 +90,8 @@ class Socket(object):
         rc = 0
         if not self._closed and hasattr(self, '_zmq_socket'):
             if self._zmq_socket is not None:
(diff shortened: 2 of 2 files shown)


dovf:
There are a couple of open issues which *may* be related to this -- #837, possibly also #845? But there are many variables (what backend is being used? where exactly is the hang occurring? etc.), and it's hard to tell whether they are the same issue or not...

The truth is that I was having some issues with linger, and when starting to dig into the code I came across the above bugs, which are quite obviously bugs; however, I don't think they're really related to the problems I was having, 'cause I doubt that I was using the cffi backend! However, the minimal test-case I used to reproduce this (the one included in the commit message) seems to be working for me with the cython backend, so at the moment I can't point to a specific problem with the cython backend...

minrk:
@dovf thanks! I didn't think to check which implementation people were using.

PR #867, opened 2016-07-23
Sync the behavior of zmq_poll between cffi and cython

### LINKED ISSUE #863: The behavior of tornado iostream connect is different between pypy and cpython.
Hello, I have encounter a strange behavior between pypy and python. 
The following is the environment for pypy and  cpython.

```
libzmq-4.1.2
pyzmq-15.2.0
Python-2.7.10 (b0a649e90b6642251fb4a765fe5b27a97b1319a9, May 05 2016, 17:21:19)
[PyPy 5.1.1 with GCC 4.9.2]
```
```
libzmq-4.1.2
pyzmq-15.2.0
Python-2.7.11 (default, May 16 2016, 11:20:19) 
[GCC 4.8.4]
```
When using tornado iostream to connect to a non exist host, cpython will wait until timeout while pypy will just yield the future. 

The following is the minial code snippets for this issue.
```python
import socket
from datetime import timedelta
from tornado import gen, iostream
# from tornado.ioloop import IOLoop
from zmq.eventloop import ioloop
IOLoop = ioloop.IOLoop


@gen.coroutine
def connect():
    try:
        host = "10.1.1.1" # or other non exist host
        port = 80
        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        stream = iostream.IOStream(sock)
        print "start connect"
        yield gen.with_timeout(timedelta(seconds=3), stream.connect((host, port)))
        print "end connect"
    except Exception as e:
        print e

_loop = IOLoop().instance()
_loop.run_sync(connect)
_loop.start()

```

The output for cpython
```
start connect
Timeout
```

The output for pypy
```
start connect
end connect
```

Another things to notice is that if I use the tornaod ioloop instead of the pyzmq one, the behavior will be what I expected.

Any help will be thankful.

chhsiao90:
This PR tried resolved #863 

It seams that there is some difference between cython poll and cffi poll that caused strange ioloop behavior when running on pypy.

That cython check event no, and ignore event = 0
But in cffi, there is some case that it would not ignore event = 0

From zmq/backend/cython/_poll.pyx:125
```python
for i in range(nsockets):
    revents = pollitems[i].revents
    # for compatibility with select.poll:
    # - only return sockets with non-zero status
    # - return the fd for plain sockets
    if revents > 0:
        if pollitems[i].socket != NULL:
            s = sockets[i][0]
        else:
            s = pollitems[i].fd
        results.append((s, revents))
```

From zmq/backend/cffi/_poll.py:43
```python
for index in range(len(items)):
    if not items[index].socket == ffi.NULL:
        if items[index].revents > 0:
            result.append((low_level_to_socket_obj[items[index].socket][0],
                        items[index].revents))
    else:
        result.append((items[index].fd, items[index].revents))
```


minrk:
Thanks!

chhsiao90:
Hi @minrk ,
Sorry I make mistake to push the swp file too.
Should I send another PR for it? thanks.

chhsiao90:
Ok, I had seen that it had bean removed.
Thanks for your support and sorry about the mistake..

PR #874, opened 2016-07-28
fix: auth thread start fail in python2.6

masuo86:
Event.wait()  fun always return NONE in python 2.6, so auth thread terminate in python verison 2.6


minrk:
Thanks!

PR #881, opened 2016-08-04
add zmq.DRAFT_API

minrk:
and don't expect draft constants to be defined unless draft API is in use

fixes builds with libzmq master (4.2), which has deleted some draft constants.

Don't move anything for an unreleased version of libzmq out of draft before a release candidate is made,
because it's always changing and deleting APIs and constants.


minrk:
appveyor failure fixed by #882

PR #890, opened 2016-08-16
Use * to refer to random port.

Kentzo:
0 is a value of INADDR_ANY and specific to the tcp transport.
This value is incorrect for VMCI, where random port is set by -1.
In other hand, '*' is used universally since ZeroMQ 2-x and therefore
should be preferred over '0'.


minrk:
Thanks!

Kentzo:
@minrk I think I forgot to fix the comment…

PR #899, opened 2016-08-25
Fix leak in implementation of RCVTIMEO + SNDTIMEO

### LINKED ISSUE #898: leak with new RCVTIMEO code
seems like there's a leak with this code.  When I use this tracemalloc lists a leak from the recv method.  I'm guessing the future has an internal reference that isn't being released.

thehesiod:
Issue is that _recv_futures + _send_futures are cleared during
_handle_recv, however if you always timeout and never receive the lists
will never get cleared and will keep on chewing up memory

closes #898


minrk:
Thanks!

Issue #903, opened 2016-08-28
Support building against bundled libzmq for cython extensions?


alexrudy:
I'd like to build a cython extension against pyzmq. In order to do this portably, I'd like to be able to link against the bundled libzmq when it is present. I think that all I need to do this is to have a portable way to get the libzmq library path (like ``get_includes()``)  and to keep the bundled libzmq's header files somewhere sensible in pyzmq when it is bundled, adding that path to the output of ``get_includes``.

I'm happy to try implementing this, but I want to make sure there is not a good reason to avoid this.

This comes up for me a lot because pip defaults to installing bundled copies of libzmq, which then conflict with a system built libzmq, but that might be a harder issue to get around. (On my own machine, I can always do ``pip install --no-use-wheel pyzmq``, but I don't always have control over how pyzmq is installed.

minrk:
You will have to be a little careful, because you are not guaranteed that libzmq is bundled, and pyzmq doesn't bundle libzmq's headers, and further *shouldn't* if libzmq is not bundled as an extension itself. So you will always have to cover both cases of bundled libzmq and non-bundled libzmq.

But in principle, adding the zmq headers when libzmq is bundled as an extension is sensible.

alexrudy:
I will ensure that pyzmq won't bundle libzmq headers when libzmq is not bundled, and I'd prefer to reliably have unbundled versions, so I still think that should be the default. But I'll add the ability to include the libzmq headers when bundled.

minrk:
Fixed by #905

PR #905, opened 2016-08-29
Bundle libzmq headers when libzmq is bundled

alexrudy:
This PR is in response to #903. When libzmq is bundled with pyzmq, this PR also bundles the libzmq header files in ``zmq/include`` so that you can compile against the bundled libzmq. It adds this header directory to ``zmq.get_includes()`` and adds a new function ``zmq.get_library_dirs`` which returns the appropriate library search path for binding against the bundled libzmq.

I've caused many bugs in my own code because I've compiled against the pyzmq cython extensions, but linked to a copy of libzmq previously installed on my machine. Ideally, the solution is to install pyzmq with ``pip install --no-binary :all: pyzmq``, but I don't always have control over how pyzmq is installed. If pyzmq has already been linked to its own copy of libzmq, including the libzmq headers for the bundled copy would allow me to build cython extensions against pyzmq.

Packaging is hard, so I suspect there are edge cases or specifics I might have missed here.

### DIFF
--- a/zmq/__init__.py
+++ b/zmq/__init__.py
@@ -46,7 +46,7 @@ def get_includes():
         includes.append(pjoin(parent, base, 'include'))
     return includes
     
-def get_libdirs():
+def get_library_dirs():
     """Get the library directory for linking to the bundled copy of libzmq."""
     from os.path import join, dirname, abspath, pardir
     base = dirname(__file__)
--- a/zmq/__init__.py
+++ b/zmq/__init__.py
@@ -38,10 +38,26 @@ from zmq.sugar import *
 
 def get_includes():
     """Return a list of directories to include for linking against pyzmq with cython."""
+    from os.path import join, dirname, abspath, pardir, exists
+    base = dirname(__file__)
+    parent = abspath(join(base, pardir))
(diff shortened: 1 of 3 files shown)


zmq/__init__.py
```diff
@@ -38,10 +38,26 @@ def _load_libzmq():
...
+    
+def get_library_dirs():
+    """Get the library directory for linking to the bundled copy of libzmq."""
     from os.path import join, dirname, abspath, pardir
     base = dirname(__file__)
     parent = abspath(join(base, pardir))
-    return [ parent ] + [ join(parent, base, subdir) for subdir in ('utils',) ]
+    global _libzmq
+    try:
+        _libzmq
```
alexrudy: Thinking about this more: This is probably the wrong way to detect bundled libzmq at runtime. What is the right way?

minrk: I don't think you need to detect it. Including the directory should be harmless if there isn't a libzmq in there.

alexrudy:
`get_library_dirs` now returns the directory that bundled libzmq would be in, regardless of whether it is present or not.

minrk:
Thanks!

PR #909, opened 2016-09-13
Fix building with recent versions of Cython

jdemeyer:
Cython 0.25 (not released yet) introduced a completely new version of `Cython.Distutils.build_ext`. See https://github.com/cython/cython/pull/1456

Since `pyzmq` uses a lot of internals of the old `build_ext`, it no longer works. However, the old implementation is still available in `Cython.Distutils.old_build_ext`.


minrk:
Thanks!

PR #913, opened 2016-09-21
send() takes at least 1 positional argument

bucketzxm:
This function may cause python exception if it doesn't have positional argument msg.


minrk:
Do you have a test that triggers this?

PR #914, opened 2016-09-22
Fix zmq.get_includes and zmq.get_library_dirs

alexrudy:
My PR #905 contained two bugs. This PR fixes those two bugs, and adds a quick test case for them.


minrk:
Great!

Issue #918, opened 2016-10-03
Handle cancelled recv


thehesiod:
if you've set a RCVTIMEO on a socket and cancel a pending recv's future, later pyzmq will try to re-set the recv's future when data is available causing an error.  pyzmq should either watch for the future being done and cancel setting it, or check before setting the future of the recv if it's aready "done"

minrk:
Do you have a test case? The RCVTIMEO code checks `future.done()`, which should be enough, but maybe there's one missing that I don't see.

thehesiod:
see 
```Python
handle: <Handle _AsyncSocket._deserialize.<locals>._chain(<Future finis...aaaaq\x04sb.'>) at /Users/amohr/dev/third_party/pyzmq/zmq/eventloop/future.py:171>
Traceback (most recent call last):
  File "/usr/local/Cellar/python3/3.5.2_2/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/events.py", line 125, in _run
    self._callback(*self._args)
  File "/Users/amohr/dev/third_party/pyzmq/zmq/eventloop/future.py", line 181, in _chain
    f.set_result(loaded)
  File "/usr/local/Cellar/python3/3.5.2_2/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/futures.py", line 329, in set_result
    raise InvalidStateError('{}: {!r}'.format(self._state, self))
asyncio.futures.InvalidStateError: CANCELLED: <Future cancelled>
```

thehesiod:
I see the issue, you're not checking for cancelled

thehesiod:
I'll submit a patch, btw, could you put something on your page comparing pyzmq to aiozmq.  What are the differences?

thehesiod:
ok, created PR which will cancel the recv future if it hasn't completed already.  This should correctly propagate cancels

PR #922, opened 2016-10-05
cancel underlying recv event if client cancels chained future

thehesiod:
fixes https://github.com/zeromq/pyzmq/issues/918


minrk:
Thanks! Went with #923 for a similar fix.

PR #941, opened 2016-11-15
[WIP] limit starvation on async sockets

minrk:
Set `async.Socket.starvation_limit = number_of_seconds` to put an upper bound on starvation caused by greedy send/recv.

Disable starvation checking with

```python
asyncio.Socket.starvation_limit = 0
```

Default is 20ms (`starvation_limit = 0.02`), which has performance impact of 2-5% on [this benchmark](https://github.com/achimnol/asyncio-zmq-benchmark).

prompted by https://github.com/dabeaz/curio/issues/106

TODO:

- [ ] performance profile
- [ ] compare with simpler count-based limits (msg or byte count?)

### DIFF
--- a/zmq/eventloop/future.py
+++ b/zmq/eventloop/future.py
@@ -5,12 +5,20 @@
 
 from collections import namedtuple
 from itertools import chain
+import weakref
+
 from zmq import POLLOUT, POLLIN
 
 try:
     from tornado.concurrent import Future
 except ImportError:
     from .minitornado.concurrent import Future
+
+try:
+    from time import monotonic as now
+except ImportError:
+    # use time.clock on Python 2
+    from time import clock as now
 
 class CancelledError(Exception):
     pass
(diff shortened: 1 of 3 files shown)


bluca:
This PR has suffered from starvation :-) Is it good to go?

minrk:
No, I'm still investigating whether it's a good idea.

dbivolaru:
In more complex applications, if it gets hairy and you have multiple sockets you do want "equal" asyncio scheduling. Sockets which have a lot of data traffic should be scheduled more often (otherwise performance is impacted), so the question is what should happen when everyone gets a lot of data.

That is why I think a size based starvation algorithm would be better for throughput based applications and a message count based starvation algorithm would be better for highly transactional applications. Adding both options so people can choose would not be bad.

The current approach for time based starvation will probably be detrimental for highly transactional applications.

PR #947, opened 2016-11-26
problem: missing Vagrantfile

wesyoung:
solution: add Vagrantfile, helps bootstrap development a bit by pre-configuring a vbox instance with all the fix'ns...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+.idea
+.vagrant
 *.pyc
 zmq/backend/cython/*.c
 zmq/devices/*.c
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,31 @@
+#e -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+# This will setup a clean Ubuntu1404 LTS env
+
+$script = <<SCRIPT
+add-apt-repository ppa:fkrull/deadsnakes-python2.7
+apt-get update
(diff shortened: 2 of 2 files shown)


minrk:
I don't quite understand what this is for. Would better documenting the dev requirements be a better solution?

wesyoung:
Vagrantfile is sorta like a Dockerfile or a Makefile, but for the vagrant toolset.

vagrant is a tool that makes it easier to bootstrap fresh virtualbox (or aws or vmware) instances where those instances are in a known working state. it goes a step further than what dev_requirements can offer, in that it installs all the system level stuff (build-essentials, gcc, python-dev, virtualenv, etc) when you're a developing across many different projects (which can have diff dep requirements).

for instance, i develop on ubuntu14 (using vagrant) on osx. and rather than having to blow away a virtual machine and rebuild from scratch manually, vagrant automates a lot of that base level work for me (running apt-get, etc). also, as a project maintainer if someone comes to me with a weird bug, we can both use a base state vm (leveraging vagrant) to help highlight where the bug might be, other than "hey, it works in my environment".

it doesn't seem like a lot, but i tend to work on more than 10-15 projects at a time [as i'm sure others do too], and the simple automation of quickly bringing up pre-configured VM's ready to go for a project drastically reduces the amount of overhead getting into a project (i understand it can be derived from the readme/dev_requirements, but automation of a build env can make all the difference towards a project imo)

see: https://www.vagrantup.com/docs/why-vagrant/

does that make sense? just trying to find ways to help others develop more pyzmq :)

minrk:
ok, thanks

PR #951, opened 2016-12-13
SUBSCRIBE/UNSUBSCRIBE notifies state changing for gevent

sublee:
`set()` with `SUBSCRIBE` and `UNSUBSCRIBE` actually sends a message like `send()`.  So this event should be notified to the I/O loop.

This patch fixes [one of crash issues](https://github.com/zeromq/libzmq/issues/2252) on `PUB`/`SUB` sockets.


minrk:
Thanks!

PR #952, opened 2016-12-18
Signal inheritance fix for zmq.devices.ProcessProxy.

dbivolaru:
# Problem Statement
zmq.devices.ProcessProxy won't react to SIGTERM (Process.terminate() or even manually do it from commandline). SIGKILL does work. Actual example from hang.
```
$ sudo netstat -antpl | grep 555
tcp        0      0 127.0.0.1:5555          0.0.0.0:*               LISTEN      11437/python3       
tcp        0      0 127.0.0.1:5556          0.0.0.0:*               LISTEN      11438/python3       
tcp        0      0 127.0.0.1:5557          0.0.0.0:*               LISTEN      11437/python3       
tcp        0      0 127.0.0.1:5558          0.0.0.0:*               LISTEN      11438/python3
$ kill -SIGTERM 11437
$ kill -SIGTERM 11438
$ sudo netstat -antpl | grep 555
tcp        0      0 127.0.0.1:5555          0.0.0.0:*               LISTEN      11437/python3       
tcp        0      0 127.0.0.1:5556          0.0.0.0:*               LISTEN      11438/python3       
tcp        0      0 127.0.0.1:5557          0.0.0.0:*               LISTEN      11437/python3       
tcp        0      0 127.0.0.1:5558          0.0.0.0:*               LISTEN      11438/python3
```

# Example code
The code running (simplified here) is:
```
import zmq
import zmq.devices

p = zmq.devices.ProcessProxy(zmq.ROUTER, zmq.DEALER)
p.bind_in('tcp://127.0.0.1:5558')
p.bind_out('tcp://127.0.0.1:5556')
p.setsockopt_in(zmq.LINGER, 0)
p.setsockopt_out(zmq.LINGER, 0)
p.start()

q = zmq.devices.ProcessProxy(zmq.XSUB, zmq.XPUB)
q.bind_in('tcp://127.0.0.1:5555')
q.bind_out('tcp://127.0.0.1:5557')
q.setsockopt_in(zmq.LINGER, 0)
q.setsockopt_out(zmq.LINGER, 0)
q.start()

# Send SIGTERM
p.launcher.terminate()
q.launcher.terminate()

# Here it waits indefinitely when ran in unittests but works when run as separate script
p.join()
q.join()
```
But this simplified version seems to work. If I comment out the code for ProxyDevice from my unittests and then run it separately where it doesn't get setUp() and tearDown() then I am unable to reproduce the error this way.

# Environment
```
libzmq-4.1.4
pyzmq-15.3.0
Python-3.5.2 (default, Oct 14 2016, 12:54:53) 
[GCC 6.2.1 20160916 (Red Hat 6.2.1-2)]
```

# Solution

It turns out the signal handling in the application (used for proper socket and context cleanup before quitting) was getting inherited with the ProxyDevice which made it impossible to stop it.

Fix is to make the call to ```start()``` with default signal handlers:
```
old_sigterm_action = signal.signal(signal.SIGTERM, signal.SIG_DFL)
old_sigint_action = signal.signal(signal.SIGINT, signal.SIG_DFL)
p.start()
q.start()
signal.signal(signal.SIGTERM, old_sigterm_action)
signal.signal(signal.SIGINT, old_sigint_action)
```

# Proposal

Update pyzmq library to make zmq.devices.ProcessProxy always react to SIGTERM and SIGINT,

What do you guys think?

### DIFF
--- a/zmq/devices/basedevice.py
+++ b/zmq/devices/basedevice.py
@@ -7,6 +7,7 @@
 import time
 from threading import Thread
 from multiprocessing import Process
+from signal import signal, SIGTERM, SIGINT, SIG_DFL
 
 from zmq import device, QUEUE, Context, ETERM, ZMQError
 
@@ -201,7 +202,14 @@ class BackgroundDevice(Device):
     def start(self):
         self.launcher = self._launch_class(target=self.run)
         self.launcher.daemon = self.daemon
-        return self.launcher.start()
+        
+        old_sigterm_action = signal(SIGTERM, SIG_DFL)
+        old_sigint_action = signal(SIGINT, SIG_DFL)
+        try:
+            self.launcher.start()
+        finally:
+            signal(SIGTERM, old_sigterm_action)
+            signal(SIGINT, old_sigint_action)
 
     def join(self, timeout=None):
         return self.launcher.join(timeout=timeout)


minrk:
I think it's important that the user be able to set the signal handling, which this seems to prevent. This also makes it impossible for a process group with ProcessDevices to disable interrupts, because the subprocesses inherit signals sent to the parent, so sending SIGINT to a process which has explicitly disabled SIGINT will stay up, but all of its devices will shut down.

I think following the standard behavior of inheriting the parent process environment makes the most sense, and cases where the context should differ can do so, as you've shown here.

Alternately, restoring the default handlers could make sense, but if we do so we must also detach it from the current process group, so that it doesn't receive its parent's signal handlers.

dbivolaru:
The ProcessDevice is a stand-alone fully working program "black box" that assumes SIGTERM and SIGINT have a certain behavior, as there is no other way to kill it. The user should not be able to change the signal handling for ProcessDevice so it's able to finalize it properly. If the user overrides the signals then there is absolutely no way to bring it down save for SIGKILL.

A side-bug of this is leaving open orphaned processes if your main script dies when signals are over-ridden which results in bind() errors when you restart your program. That is why detaching from current process group is also not good.

minrk:
It is important to allow users to make signaling choices. It is perfectly reasonable to have a parent process capture SIGINT for patterns such as 'confirm exit', etc. If the ProcessDevice unconditionally exits when the parent receives SIGINT, this would be very bad. Right now, users can implement this by setting signal handlers prior to fork. As far as I can tell, this PR would make this pattern impossible. The parent being able to capture SIGINT and prevent it from propagating to the child is not a bug.

dbivolaru:
I understand the issue you are talking about. I wrote the following mini-test to illustrate all the possible combinations.

What you are saying is that Test Case 1a should be FAIL. But in my opinion it's better to receive an Exception (feedback) than to have a silently running ProcessDevice de-attached to anything that keeps your ports reserved as in Test Case 2a. That is hard(er) to troubleshoot than just getting the nice exception.

But you do have a point with the fact that in 1a type of scenarios, one might need a controlled stop where the Proxy is the last one going down (and not right away) to allow for all messages to be sent around. I'll have a look at your idea to detach it from the process group.

Test Case | main | ProcessDevice | Results | Pass/Fail
---|---|---|---|---
1a CTRL+C pressed | SIG_IGN | SIG_DFL | ProcessDevice dies with KeyboardInterrupt; main keeps running until end | PASS
1b CTRL+C not pressed | SIG_IGN | SIG_DFL | ProcessDevice and main keep running until end | PASS
2a CTRL+C pressed | SIG_DFL | SIG_IGN | main quits abruptly but ProcessDevice keeps running in background | FAIL
2b CTRL+C not pressed | SIG_DFL | SIG_IGN | main never finishes as it hangs on join() | FAIL

```
import zmq
import zmq.devices
import time
import signal

# Test Cases
# 1 - main process: SIG_IGN, ProcessDevice: SIG_DFL
#       a - press CTRL+C while sleep()'ing
#       b - wait for script to finish
# 2 - main process: SIG_DFL, ProcessDevice: SIG_IGN
#       a - press CTRL+C while sleep()'ing
#       b - wait for script to finish
# 3 - main process: SIG_IGN, ProcessDevice: SIG_IGN (obvious)
# 4 - main process: SIG_DFL, ProcessDevice: SIG_DFL (obvious)
test_case = 2

p = zmq.devices.ProcessProxy(zmq.ROUTER, zmq.DEALER)
p.bind_in('tcp://127.0.0.1:5558')
p.bind_out('tcp://127.0.0.1:5556')
p.setsockopt_in(zmq.LINGER, 0)
p.setsockopt_out(zmq.LINGER, 0)

if test_case in [2, 3]:
    signal.signal(signal.SIGTERM, signal.SIG_IGN)
    signal.signal(signal.SIGINT, signal.SIG_IGN)
p.start() # ProcessDevices inherits signal handler at creation time

# Main process ignores
if test_case in [1, 3]:
    signal.signal(signal.SIGTERM, signal.SIG_IGN)
    signal.signal(signal.SIGINT, signal.SIG_IGN)
else:
    signal.signal(signal.SIGTERM, signal.SIG_DFL)
    signal.signal(signal.SIGINT, signal.SIG_DFL)
time.sleep(60)

p.launcher.terminate()
p.join()
print('done')
```

minrk:
Thanks for the test case!

> What you are saying is that Test Case 1a should be FAIL

Yes, I would consider that a failure to do the intended/expected action. It's not hypothetical, either; it's something that I rely upon on a regular basis. It is important that users be able to prevent SIGINT to the parent from halting the child, ideally in the simple, standard way that works for all subprocesses spawned from Python. Taking away valid user choices is something I prefer to avoid.

This PR does reduce the negative consequences of certain non-default choices (i.e. registering signal handlers without including proper cleanup code in the parent), but it makes important, valid use cases unavailable.

dbivolaru:
The point with user choice is very valid. With SIG_IGN (or other inherited signal handler assuming a controlled stop scenario), the are not so many choices, albeit ugly one has full control in the main script to really only send the kill signal to the ProcesDevice really at the very end.

```os.kill(p.launcher.pid, signal.SIGKILL)```

instead of

```p.terminate()```

Issue #950, opened 2016-12-13
Crash from green PUB socket on heavy workload


sublee:
I have a crash issue with green `PUB` socket.  I reported the issue at https://github.com/zeromq/libzmq/issues/2252.  I got an advice.  He guessed I'm using zmq with multi-threading.

When I use `zmq.green`, do I need to always acquire a lock for every green zmq sockets?  So I have to change all code like this:

```python
socket = context.socket(zmq.SUB)
socket.connect('tcp://127.0.0.1:5932')
socket.set(zmq.SUBSCRIBE, my_topic)
```

To like this?

```python
zmq_lock = gevent.lock.Semaphore()
with zmq_lock:
    socket = context.socket(zmq.SUB)
    socket.connect('tcp://127.0.0.1:5932')
    socket.set(zmq.SUBSCRIBE, my_topic)
```

I hope your help.

minrk:
closed by #951

sublee:
@minrk Let's talk on https://github.com/zeromq/libzmq/issues/2252.

sublee:
But please reopen this issue.  I still have problems.

sublee:
My team has tried to reproduce this crash by only libzmq but we didn't succeed yet.  Now I guess PyZMQ caused the crash rather than libzmq.

@minrk would you run the below code which reproduces the crash?

```python
# -*- coding: utf-8 -*-
from collections import defaultdict
import random
import time

import gevent.monkey; gevent.monkey.patch_time()  # noqa
import zmq.green as zmq


# Number of SUB sockets at least 2.
n = 2
# Number of topics.
m = 1000


t = time.time()
c = zmq.Context()
p = c.socket(zmq.PUB)
p.set(zmq.RCVHWM, 1000)  # It doesn't matter to lead the crash.
topic = lambda x: ('%08x' % abs(hash(str(x))))[:8]


print '1. Spawn %d SUB sockets' % n
subs = []
ports = []
for x in range(n):
    s = c.socket(zmq.SUB)
    s.set(zmq.SNDHWM, 10)  # Lower high water mark leads the crash faster.
    for x in range(m):
        s.set(zmq.SUBSCRIBE, topic(x))
    port = s.bind_to_random_port('tcp://127.0.0.1')
    ports.append(port)
    subs.append(s)


print '2. Connect to SUB sockets'
for port in ports:
    p.connect('tcp://127.0.0.1:%d' % port)


print '3. Subscribing/Unsubscribing...'
[...]

PR #964, opened 2017-01-11
[WIP] Don't call getsockopt from the gevent I/O loop

sublee:
I split `__state_changed` to `__readable_detected` to safen a callback of the gevent I/O loop.  When a callback calls `getsockopt` during an application calls `setsockopt`, it could cause the crash I reported at https://github.com/zeromq/pyzmq/issues/950.

You can reproduce the crash with `test_sigabrt_issue_case2` by reverting 847a28f.

### DIFF
--- a/zmq/tests/test_crash.py
+++ b/zmq/tests/test_crash.py
@@ -1,0 +1,100 @@
+# Copyright (C) PyZMQ Developers
+# Distributed under the terms of the Modified BSD License.
+
+
+import functools
+import hashlib
+from multiprocessing import Pipe, Process
+from random import Random
+import time
+
+import pytest
+
+import zmq
+
+from zmq.tests import BaseZMQTestCase, GreenTest, have_gevent
+
+
+def topic(x):
+    """Generates a PUB/SUB topic from a number."""
+    return hashlib.md5(str(x)).hexdigest()[:8]
+
+
+def assert_exit_code(exit_code):
+    if exit_code == 1:
+        raise AssertionError('Program error')
+    elif exit_code == -6:
+        raise AssertionError('Crashed by SIGABRT')
+    elif exit_code == -11:
+        raise AssertionError('Crashed by SIGSEGV')
+    elif exit_code < 0:
+        raise AssertionError('Crashed with exit code %d' % exit_code)
+
+
+def capture_crash(f):
+    @functools.wraps(f)
+    def wrapped(*args, **kwargs):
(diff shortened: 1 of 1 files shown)


minrk:
Thanks! The new test is crashing on Python 2.7. I'm not sure what the cause could be.

sublee:
I'm also feeling weird.  I should spend more time to explain that Python crash.

sublee:
@minrk It looks like a bug before libzmq-4.2.  When I tested it with 4.2, there was no problem.  But with 4.1.6, the bundled, it was crashed by `malloc(): smallbin double linked list corrupted: 0x0000000000f831c0`.  Do you have a plan to upgrade the bundled libzmq version to 4.2.0?

minrk:
> Do you have a plan to upgrade the bundled libzmq version to 4.2.0?

To 4.2, yes, though bugs in 4.2.0 prevent me from doing it. I've been waiting for 4.2.1 to consider it.

It seems like there is still a problem in the code if it's able to crash 4.1.6, because pyzmq will still be used with 4.1 even after 4.2 is the bundled version.

sublee:
I'm still digging this issue.  I didn't find the answer yet but I found another crash case.  @minrk you can check it at [`test_sigabrt_issue_case3`](https://github.com/what-studio/pyzmq/blob/d4d39bd2a1a8834ccc090b9da25caa06cf37190c/zmq/tests/test_pubsub.py#L131).  I wish if you also would dig this issue with me.

sublee:
I've developed crash test cases more.  Now I can run that cases with `zmq.green` or vanilla `zmq`.

I get "erased == 1 assertion failure" with `zmq.green` as you know.  But with vanilla `zmq`, they have still crash with "segmentation fault".

| Test case | Green | Outcome |
| ---------- | ------ | ----------- |
| `test_connect_and_close` | ✔️ | `Assertion failed: erased == 1 (bundled/zeromq/src/mtrie.cpp:302)` |
| `test_connect_and_close` | ✖️ | `segmentation fault (core dumped)` |
| `test_concurrent_subs` | ✔️ | `Assertion failed: erased == 1 (bundled/zeromq/src/mtrie.cpp:302)` |
| `test_concurrent_subs` | ✖️ | `segmentation fault (core dumped)` |
| `test_separate_process` | ✔️ | `Assertion failed: erased == 1 (bundled/zeromq/src/mtrie.cpp:302)` |
| `test_separate_process` | ✖️ | `AssertionError: SUB socket did not receive any message` (not crashed) |

PR #966, opened 2017-01-11
Faster no-copy by pre-creating PUSH socket

pitrou:
This improves performance by 30% on the following micro-benchmark:
```
$ python -m timeit -s "import zmq; data=b'foo'" "f = zmq.Frame(data); del f"
```
* before: 16 usec per loop
* after: 11.2 usec per loop

The cffi backend appears to always copy, and so is not affected.

### DIFF
--- a/zmq/backend/cython/message.pyx
+++ b/zmq/backend/cython/message.pyx
@@ -59,7 +59,7 @@ from zmq.utils.strtypes import bytes,unicode,basestring
 #-----------------------------------------------------------------------------
 
 ctypedef struct zhint:
-    void *ctx
+    void *sock
     size_t id
 
 cdef void free_python_msg(void *data, void *vhint) nogil:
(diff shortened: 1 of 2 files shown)


minrk:
zmq sockets are not threadsafe, so we cannot share this socket across threads, and there are one or more IO threads per context. We would need to find a way to ensure that the socket is unique per IO thread. This is knowable per-context for the common case of `io_threads=1`, but not in general, so selecting which socket to use has to happen in the `free_fn` itself. I've wanted to do this, but need a simple way to get a thread identity that works on posix and Windows.

minrk:
With zmq >= 4.2, there is the introduction of threadsafe CLIENT-SERVER sockets, which would allow us to do it this way.

I do need to check one thing, though: while zmq sockets are officially "not threadsafe", it's possible that confining inter-thread usage to single-frame send calls may actually be safe. I'm not sure if I can verify that rigorously, and I know that it's not guaranteed by the spec.

pitrou:
> I've wanted to do this, but need a simple way to get a thread identity in C that works on posix and Windows.

In that case, you can use the CPython API `long PyThread_get_thread_ident(void)`. It's not documented, but hasn't changed for a long time.
(or you can just mimick that function's implementation, i.e. call `pthread_self` on POSIX and `GetCurrentThreadId` on Windows)

bluca:
@minrk even with a single send it would still be using the same pipe from different threads, so race conditions can still happen without full memory barriers if not mutual exclusion

client-server are thread safe, but that comes with a cost so it might actually make performance worse

minrk:
I added a mutex in #967 on top of this, which seems simpler than looking up sockets by thread id, which would need its own threadsafe hashmap and complicate cleanup, etc.

PR #967, opened 2017-01-11
add mutex to ensure threadsafety in gc push

minrk:
follow-up for #966

appears to have very little effect on performance

- this branch: 6.72 us
- #966: 6.63 us
- master: 7.77 us

Haven't tested if the Windows version actually works, yet.


minrk:
Turns out windows failure wasn't due to the code in this PR, it was already failing and fixed by #975.

PR #979, opened 2017-02-02
Make setup test for "sys/un.h" cross-compile-friendly.

simias:
The previous test used compile_and_run but running the test binary on
the build host is not always possible, for instance when
cross-compiling. Furthermore it's not necessary in this case since a
missing "sys/un.h" header would be caught at compile time and the lack
of library support at link time.

Therefore this patch replaces compile_and_run with compile_and_forget
which builds the test program but doesn't run it. compile_and_run is
removed since it's now unused.

### DIFF
--- a/buildutils/detect.py
+++ b/buildutils/detect.py
@@ -58,7 +58,11 @@ def test_compilation(cfile, compiler=None, **compiler_attrs):
     cc.link_executable(objs, efile, extra_preargs=lpreargs)
     return efile
 
-def compile_and_run(basedir, src, compiler=None, **compiler_attrs):
+def compile_and_forget(basedir, src, compiler=None, **compiler_attrs):
+    """Make sure src compiles and links successfully.
+
+    The resulting binary is deleted without being run.
+    """
     if not os.path.exists(basedir):
         os.makedirs(basedir)
     cfile = pjoin(basedir, os.path.basename(src))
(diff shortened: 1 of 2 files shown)


buildutils/detect.py
```diff
@@ -58,26 +58,21 @@ def test_compilation(cfile, compiler=None, **compiler_attrs):
...
+    The resulting binary is deleted without being run.
+    """
     if not os.path.exists(basedir):
         os.makedirs(basedir)
     cfile = pjoin(basedir, os.path.basename(src))
     shutil.copy(src, cfile)
     try:
         cc = get_compiler(compiler, **compiler_attrs)
         efile = test_compilation(cfile, compiler=cc)
-        patch_lib_paths(efile, cc.library_dirs)
```
minrk: If these are a problem for cross-compiling, they should *only* be skipped when cross compiling. Under normal circumstances, these steps should still run.

simias: They're never needed though: you don't need to run the code to test for "sys/un.h", it's purely a compile thing.

IIRC previously `compile_and_run` was used to detect the ZMQ library version, but now `detect_zmq` calls `test_compilation` directly. I think this code is simply a leftover from that time, it was never needed for the sys/un test.

minrk: Good call, thanks.

simias:
Thank you!

PR #987, opened 2017-02-26
add send/recv_serialized

minrk:
generic hook for custom serialization functions

- [x] test
- [x] test with futures
- [x] docs

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -12,6 +12,8 @@ For a full changelog, consult the `git log <https://github.com/zeromq/pyzmq/comm
 17.0.0
 ======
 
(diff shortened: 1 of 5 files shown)


mrvn:
How do these function make any sense?

'socket.send_serialized(msg, serialize, flags, copy)' can just be typed as 'socket.send_multipart(serialize(msg), flags, copy)'.

'socket.recv_serialized(deserialize, flags, copy)' can just be typed as 'deserialize(socket.recv_multipart(flags, copy)'

I don't see how this adds any new feature or more comfort. Further if you have a custom serializer and deserializer you probably want to use it for every message. So every call to send/recv will have the same callbacks passed again and again.

Why not have functions to set a serializer/deserializer in the socket and then have them used automatically every time instead of passing them as callbacks?

PR #997, opened 2017-03-21
update timeout when poll is retried due to interrupted syscalls

### LINKED ISSUE #996: poller.poll() no longer returns when interrupted by signal and restarts timeout
The code below polls for activity on a socket for 5 seconds and sets up an alarm to interrupt it after 1 second. What used to happen was that poller.poll() would raise zmq.AGAIN after the signal handler returned. With libzmq 4.2.2 and pyzmq 16.0.2 poller.poll() restarts the poll with the initial timeout. The effect of this is that poller.poll() will never timeout in this example.

    #!/usr/bin/python3

    import logging
    import signal
    import zmq

    # enable logging output
    numeric_log_level = getattr(logging, "INFO", None)
    logging.basicConfig(level=numeric_log_level)

    # set up a alarm heartbeat once a second
    def on_alarm(signalnum, frame):
        logging.info("on_alarm()")
        signal.alarm(1)
    signal.signal(signal.SIGALRM, on_alarm)
    signal.alarm(1)

    # create and connect socket
    ctx = zmq.Context()
    socket = ctx.socket(zmq.DEALER)
    socket.connect("inproc://test")
    poller = zmq.Poller()
    poller.register(socket, zmq.POLLIN)

    # poll socket for 5 seconds
    logging.info("polling for 5 seconds")
    poller.poll(5000);
    logging.info("polling done")

    # cleanup
    socket.close()
    ctx.destroy()

minrk:
avoids extending timeout when interrupts arrive during timeout

closes #996

### DIFF
--- a/zmq/backend/cffi/_poll.py
+++ b/zmq/backend/cffi/_poll.py
@@ -8,6 +8,7 @@ try:
     from time import monotonic
 except ImportError:
     from time import clock as monotonic
+import warnings
 
 from ._cffi import C, ffi
 from zmq.error import InterruptedSystemCall, _check_rc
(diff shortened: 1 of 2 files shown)


mrvn:
In the loop you take the current time, poll, measure how time has passed and reduce the timeout by that amount. This has some minor problems:
1) You take the time twice every loop causing an extra syscall per loop.
2) The time between the second measurement and the first of the next loop is not accounted for causing drift. If the kernel happens to schedule right there that could be a significant amount.
3) timer have a finite resolution causing a measurement error in each loop which add up and cause more drift.
It would be better to get the start time before the loop, preserve the original timeout and compute the next loops timeout as `timeout - (now - start)` even if that makes detecting clock shifts harder.

PR #1001, opened 2017-04-03
WIP: bump bundled libzmq to 4.2.2 and enable draft by default

minrk:


sublee:
I love it!

minrk:
I'll add the extra methods for the draft sockets that need it in a subsequent PR.

bluca:
In CZMQ we have a similar problem, and the solution we use is to have DRAFT flags there too. So if a user enables it, it's at their risk of breakages. And default use cases are not exposed.

minrk:
Closing in favor of #1057, which enables draft support but not by default, and doesn't bump bundled libzmq to 4.2.2, since it doesn't build correctly on Windows.

PR #1021, opened 2017-06-01
Remove internal zmq_poll implementations for eventloop integration

### LINKED ISSUE #894: Loop-less zmq.asyncio
Hi guys,

Is it possible to re-work the `zmq.asyncio` package and remove the need of usage of `zmq.asyncio.ZMQEventLoop`?

Say if I use zmq in my [aiohttp](https://github.com/KeepSafe/aiohttp/) application in combination with [uvloop](https://github.com/MagicStack/uvloop) I can't just import and use `zmq.asyncio` easily. So that this forces me to use third-party libraries like [aiozmq](https://github.com/aio-libs/aiozmq) instead of the native [pyzmq](https://github.com/zeromq/pyzmq/).

Also please take a look at https://github.com/KeepSafe/aiohttp/issues/1092#issuecomment-241133022 describing specific issues with `zmq.asyncio.ZMQEventLoop`.

I'm not proficient in this particular area therefore I delegate the next discussion to @asvetlov and other guys who know this area well.
If someone is able to work on this, it would be really nice! And as @asvetlov said he can point on required changes.

minrk:
Instead, use edge-triggered zmq.FD.

This enables integration with any eventloop supported by asyncio, tornado, etc. (closes #894)

After experience of numerous bugs in zmq.green, this can result in hangs if not done super carefully, but it seems worth it at this point with the proliferation of asyncio implementations.

- [x] handle user retrieving `zmq.EVENTS`, which can consume events on the FD.
- [x] update docs
- [x] update examples
- [x] verify that it doesn't break anything in IPython/Jupyter.


f0t0n:
Thank you @minrk! 👍

thehesiod:
yay, this is yuuge :)

Issue #973, opened 2017-01-26
Numerous crashes when running the test suite on windows for python 3.6


cournape:
I am seeing crashes when running the test suite on python 3.6 for pyzmq 16.0.2.

While I can reproduce the crashes almost every time, unfortunately, the crash location is not always the same. I have been able to reproduce the crash fairly regularly in test_context:test_destroy if I only run tests for test_auth.py and test_context.py.

I see that appveyor does not run tests at all ATM. Would you be interested in PRs that blacklist test failing on windows ?

minrk:
> I see that appveyor does not run tests at all ATM. Would you be interested in PRs that blacklist test failing on windows ?

Absolutely! Maybe start with just context and socket tests.

PR #1056, opened 2017-08-14
Update appveyor testing

itziakos:
This PR updates the current appveyor test setup to run all the tests on Windows

in more details:

- Skip tests which are not supported on windows (e.g. using file handles for sockets).
- Skip `TestContext.test_destroy` because it segfaults
- Fix Skip behavior in `TestEINTRSysCall`.  `Pytest` cannot detect items marked as skip if they are not test methods.
- Add/increase timeouts or time.sleep() where necessary to consider slow VM builds.
- Run the `test_win32_shim` test on a separate/new console to avoid issues with the CTRL-C event been sent to all the processes which share the console and abort the appveyor build prematurely.
- Cache wheels (minor build speedups)

The builds are still a little flaky and will need to be rerun from time to time. 
This PR should address #973

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -42,9 +42,6 @@ cache:
   - C:\Users\appveyor\AppData\Local\pip\Cache
 
 install:
-  # Set the error action preference to so that output to stderr does not
-  # fail the build
-  - ps: $ErrorActionPreference = "SilentlyContinue"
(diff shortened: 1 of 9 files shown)


itziakos:
Need to remove the changes to the empty lines. Sorry for that.

minrk:
Thanks!

PR #1064, opened 2017-08-25
Drop GIL while generating crypto keys

njsmith:
Not tested; let's see what CI says :-)


njsmith:
Well, it almost passed. The one failure is in `test_recv_json_cancelled` on py34 on win-64, which looks like a flaky test to me?

minrk:
Indeed. Running all the Windows tests is a new thing, and probably has some fiddling to do.

PR #1073, opened 2017-09-06
Remove `buffer` because we no longer support Python 2.6.  Use `memoryview` instead.

cclauss:
`buffer` was removed in Python 3.  `memoryview` is supported in Python 2.7+
* https://docs.python.org/2/library/stdtypes.html#memoryview-type

### DIFF
--- a/zmq/tests/test_message.py
+++ b/zmq/tests/test_message.py
@@ -274,12 +274,8 @@ class TestFrame(BaseZMQTestCase):
             shape = shapes[:i]
             A = numpy.random.random(shape)
             m = zmq.Frame(A)
-            if memoryview.__name__ == 'buffer':
-                self.assertEqual(A.data, m.buffer)
-                B = numpy.frombuffer(m.buffer,dtype=A.dtype).reshape(A.shape)
-            else:
-                self.assertEqual(memoryview(A), m.buffer)
-                B = numpy.array(m.buffer,dtype=A.dtype).reshape(A.shape)
+            self.assertEqual(memoryview(A), m.buffer)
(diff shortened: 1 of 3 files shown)


zmq/tests/test_message.py
```diff
@@ -279,7 +274,7 @@ def test_buffer_numpy(self):
             shape = shapes[:i]
             A = numpy.random.random(shape)
             m = zmq.Frame(A)
-            if view.__name__ == 'buffer':
+            if memoryview.__name__ == 'buffer':
```
minrk: This block can be removed

zmq/tests/test_message.py
```diff
@@ -310,14 +305,14 @@ def test_noncopying_recv(self):
...
-            # buf = view(m)
+            # buf = memoryview(m)
             buf = m.buffer
             del m
             for i in range(5):
                 ff=b'\xff'*(40 + i*10)
                 sb.send(ff, copy=False)
                 m2 = sa.recv(copy=False)
-                if view.__name__ == 'buffer':
+                if memoryview.__name__ == 'buffer':
```
minrk: Same here - this will never be true

zmq/tests/test_message.py
cclauss: I wrote on that commit: _There is opportunity to remove tests but let's make sure that these changes pass the current tests before removing any of them._

zmq/tests/test_message.py
cclauss: I wrote on that commit: _There is opportunity to remove tests but let's make sure that these changes pass the current tests before removing any of them._

PR #1080, opened 2017-09-07
 Python 3.6 support and Python 3.3 EOL

cclauss:
I do not know if this is they way that you want to go but some food for thought...

* Add a PyPI trove classifier for Python 3.6.
* Python 3.3 goes EOL in a few weeks https://docs.python.org/devguide/index.html#branchstatus  Should we remove it?
* Clean out some Python 3.2 requirements.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,19 @@
 language: python
-dist: trusty
 
 cache:
   - apt
   - pip
+
 python:
   - 2.7
   - 3.6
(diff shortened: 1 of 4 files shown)


minrk:
> Should we remove it?

It's good to know that it's EOL and we can remove it once maintaining support is a problem. But in general, I don't like dropping support for things before it causes a maintenance burden.

cclauss:
Closing in favor of #1079

PR #1110, opened 2017-11-19
Clean up some documentation, particularly around the NOBLOCK flag.

### LINKED ISSUE #1101: pyzmq NOBLOCK recv() quite slow due to exception overhead
Hey, when zmq.Socket.recv() (and recv_multipart(), etc) is used in non blocking mode (flags=NOBLOCK), failure to read a message results in an exception being thrown (zmq.error.Again).

This is a little funny/nonpythonic because this isn't really an error condition, it's actually a very nominal condition with non-blocking I/O.

More significantly, however, throwing that exception is quite costly. On the Raspberry Pi I'm working on, for example, it ends up taking north of one millisecond to construct that exception, populate its stack trace (which means stat()'ing the backend module, etc), and generally perform related bookkeeping -- even though I'm just going to catch & throw away that exception and return to a polling loop.

That means if I have a 10msec polling loop (as part of a control system), and it checks zmq every time through that loop, pyzmq ends up burning *over 10% of the CPU* even when no messages are actually arriving.

It would be strongly desired to have some flag in the API which would allow non-blocking receive without throwing an exception when no messages are pending. (Or, equivalently, check to see if any messages are pending without performing an actual receive, or throwing an exception.)

egnor:
closes #1101

### DIFF
--- a/zmq/backend/cython/socket.pyx
+++ b/zmq/backend/cython/socket.pyx
@@ -725,11 +725,13 @@ cdef class Socket:
     #-------------------------------------------------------------------------
 
     cpdef send(self, object data, int flags=0, copy=True, track=False):
-        """s.send(data, flags=0, copy=True, track=False)
-
-        Send a message on this socket.
+        """Send a single zmq message frame on this socket.
 
         This queues the message to be sent by the IO thread at a later time.
+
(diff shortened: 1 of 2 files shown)


minrk:
Thanks!

PR #1141, opened 2018-02-12
Test PUB/SUB crash

sublee:
Hi, long time no talk.

Do you remember https://github.com/zeromq/libzmq/issues/2252, https://github.com/zeromq/pyzmq/issues/950, and https://github.com/zeromq/pyzmq/pull/964?  I'm investing into this `SIGABRT` issue again.  Because it hurt my production servers.

```
Assertion failed: erased == 1 (src/mtrie.cpp:297)
```

Actually, we already know how do we fix this issue since the last PR.  Calling `getsockopt()` in the gevent I/O callback indicates the crash.  But when I sent the last PR, I failed to make simple and focused test cases.  So we gave up to merge it.

Today, I made only 1 test case which reproduces the `SIGABRT` what I'm having.  I marked this test case with `pytest.mark.xfail`.  `XPASS` means the test case is passed while `xfail` means the test case is failed but it was expected.  You can see a test result at https://travis-ci.org/what-studio/pyzmq/jobs/340322543:

```
zmq/tests/test_crash.py::TestPubSubCrash::test_pubsub_crash XPASS        [ 20%]
zmq/tests/test_crash.py::TestPubSubCrashGreen::test_pubsub_crash xfail   [ 20%]
```

As you saw, the crash happens only with gevent.

If you accept this PR, I'll write a new PR to fix this issue.


minrk:
Careful with xfail, since it means the tests are not run at all, it's just there for your own record keeping. When I removed xfail, the test didn't run because of unicode issues on Python 3.

zmq/tests/test_crash.py
```diff
@@ -0,0 +1,100 @@
...
+import pytest
+
+import zmq
+
+from zmq.tests import BaseZMQTestCase, GreenTest, have_gevent
+
+
+def topic(x):
+    """Generates a PUB/SUB topic from a number."""
+    return hashlib.md5(str(x)).hexdigest()[:8]
```
minrk: A simpler way to create a topic from a number is to use the number itself: `return struct.pack('Q', x)`

zmq/tests/test_crash.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def capture_crash(f):
+    @functools.wraps(f)
+    def wrapped(*args, **kwargs):
+        p = Process(target=f, args=args, kwargs=kwargs)
+        p.start()
+        p.join()
+        if p.exitcode == 1:
+            raise AssertionError('Test failed')
```
minrk: This is redundant with checks in assert_exit_code

sublee:
Hey, thank you for your review!  But for now, I should pause to research this issue in PyZMQ.  I reproduced this crash without PyZMQ, but only libzmq-4.2.3.

I found the crash point at calling `getsockopt()` with `ZMQ_EVENTS`, which should be always called in `__state_changed()` for gevent-awareness.

You will want to see https://github.com/zeromq/libzmq/issues/2942!

sublee:
I think I fixed this issue in PyZMQ by calling `select()` instead of `getsockopt(EVENTS)`.  How do you think?

zmq/tests/test_crash.py
sublee: The topics should not share a common prefix.  Hashing is necessary to reproduce this issue.

sublee:
I fixed this issue by `select()` instead of `ZMQ_EVENTS`.  Please review it again.  Thank you!

zmq/green/core.py
```diff
@@ -95,30 +95,36 @@ def __setup_events(self):
...
-            self.__readable.set_exception(exc)
-        else:
-            if events & zmq.POLLOUT:
-                self.__writable.set()
-            if events & zmq.POLLIN:
-                self.__readable.set()
+        # getsockopt(ZMQ_EVENTS) at here can cause SIGABRT
+        # https://github.com/zeromq/libzmq/issues/2942
+        fd = self.getsockopt(zmq.FD)
+        readable, writable, __ = select([fd], [fd], [], timeout=0)
```
minrk: This is not correct. The zmq FD is only ever readable, never writable. The FD being readable only tells you that there *might* be read or write events waiting to process. The readable and writable events can only ever be set by interrogating ZMQ_EVENTS.

zmq/tests/test_crash.py
minrk: That's not accurate. This failure is reproducible using struct.pack. They may have a common prefix, but the topics are still at least as unique (perfectly unique in the struct.pack case, almost certainly unique in the md5 case). The same is true of md5, where 1/16 start with 'a', etc.

zmq/green/core.py
sublee: Oh, okay.  I didn't know about the difference between `ZMQ_EVENTS` and `select()`.

It seems to be the only way to fix the crash that we fix the original issue from libzmq.

sublee:
I understood what happened.

A `SUB` socket can drop `SUBSCRIBE` due to the `HWM` limit.  But the `SUB` socket still can send `UNSUBSCRIBE` of the dropped prefix.  The connected `PUB` socket receives the `UNSUBSCRIBE` but it has no pipe to remove because it has not received `SUBSCRIBE` for the prefix.  At that time, the `PUB` socket fails with the assertion, the application crashes with `SIGABRT`.

We can set `SNDHWM=0` for `SUB` and `RCVHWM=0` for `PUB`, which means keeping all subscription messages.

My patch was not the answer.  So I reverted all code changes.  But the crash test cases still should be tracked.  I hope you accept the test cases only.

sublee:
Now I think calling `self.__state_changed()` in `_Socket.set()` is not necessary to avoid any crash.  So I want to rollback it here.

minrk:
Thanks for tracking down the issue! If I understand your explanation correctly, it's a pure libzmq issue that pyzmq can't do anything about? And it's mostly a coincidence that it is hard to reproduce without zmq.green? If so, I'm reticent to include a crash test that doesn't cover anything related to pyzmq. Especially one that will fail when libzmq fixes their bug. I think attaching the test case to libzmq itself is probably a better solution, unless there really is something that we are doing wrong in zmq.green.

sublee:
@minrk I agree with you.  I'll close this PR but I want to open a new one.  That will revert my previous patch #951.

Now I think calling `self.__state_changed()` on `zmq.SUBSCRIBE` or `zmq.UNSUBSCRIBE` actually didn't fix any problem.  Because the previous crash issue was exactly same with this issue.  `test_sigabrt_issue` what I made won't fail if we config unlimited HWMs.  Do you agree with me?

minrk:
I think that's probably right

sublee:
Thank you!  I'll do it.

PR #1142, opened 2018-02-14
Revert PR #951

sublee:
We've talked about it at https://github.com/zeromq/pyzmq/pull/1141.

There is a crash issue on PUB/SUB sockets.  The bug is at libzmq not PyZMQ.  It is reported at https://github.com/zeromq/libzmq/issues/2942.  When I worked on https://github.com/zeromq/pyzmq/pull/951, I thought it is PyZMQ's thread-unsafety issue.  But it seems to be wrong.

I want to revert my code to make PyZMQ focus on only its problems.


minrk:
Thanks!

PR #1152, opened 2018-03-20
bump bundled libzmq to 4.2.5

minrk:
hopefully this fixes the Windows issues that prevented bundling 4.2.2

- [x] wait for #1153 to fix tornado tests

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,6 +11,12 @@ environment:
     CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\tools\\run_with_env.cmd"
 
   matrix:
+    - PYTHON: "C:\\Python27"
+      PYTHON_VERSION: "2.7.x"
+      PYTHON_ARCH: 32
+    - PYTHON: "C:\\Python27-x64"
+      PYTHON_VERSION: "2.7.x"
+      PYTHON_ARCH: 64
     - PYTHON: "C:\\Python36"
       PYTHON_VERSION: "3.6.x"
       PYTHON_ARCH: 32
@@ -22,12 +28,6 @@ environment:
       PYTHON_ARCH: 32
     - PYTHON: "C:\\Python35-x64"
       PYTHON_VERSION: "3.5.x"
-      PYTHON_ARCH: 64
-    - PYTHON: "C:\\Python27"
-      PYTHON_VERSION: "2.7.x"
-      PYTHON_ARCH: 32
-    - PYTHON: "C:\\Python27-x64"
-      PYTHON_VERSION: "2.7.x"
       PYTHON_ARCH: 64
     - PYTHON: "C:\\Python34"
       PYTHON_VERSION: "3.4.x"
--- a/buildutils/bundle.py
+++ b/buildutils/bundle.py
@@ -32,14 +32,14 @@ pjoin = os.path.join
 # Constants
 #-----------------------------------------------------------------------------
 
-bundled_version = (4, 2, 3)
+bundled_version = (4, 2, 4)
 vs = '%i.%i.%i' % bundled_version
 libzmq = "zeromq-%s.tar.gz" % vs
 libzmq_url = "https://github.com/zeromq/libzmq/releases/download/v{vs}/{libzmq}".format(
     vs=vs,
     libzmq=libzmq,
 )
-libzmq_checksum = "sha256:8f1e2b2aade4dbfde98d82366d61baef2f62e812530160d2e6d0a5bb24e40bc0"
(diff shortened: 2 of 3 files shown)


bluca:
@sigiesec this is nae good:

```
self = <zmq.tests.test_auth.TestThreadAuthentication testMethod=test_null>
    def test_null(self):
        """threaded auth - NULL"""
        # A default NULL connection should always succeed, and not
        # go through our authentication infrastructure at all.
        self.auth.stop()
        self.auth = None
        # use a new context, so ZAP isn't inherited
        self.context = self.Context()
    
        server = self.socket(zmq.PUSH)
        client = self.socket(zmq.PULL)
        self.assertTrue(self.can_connect(server, client))
    
        # By setting a domain we switch on authentication for NULL sockets,
        # though no policies are configured yet. The client connection
        # should still be allowed.
        server = self.socket(zmq.PUSH)
        server.zap_domain = b'global'
        client = self.socket(zmq.PULL)
>       self.assertTrue(self.can_connect(server, client))
E       AssertionError: False is not true
```

Do you remember what the problem was there with the ZAP_DOMAIN and the NULL auth? I can't find it again but pretty sure we've been through that before

bluca:
Something similar to this perhaps? https://github.com/zeromq/libzmq/issues/2762
I might have missed something when wrapping them up in the new socket option: https://github.com/zeromq/libzmq/pull/2773

Looks like a regression to me :-/

chrisjbillington:
Just FYI zmq 4.2.4 is out with some fixes for Windows, these fixes may address reasons why Anaconda devs were hesitant to update the zeromq used by pyzmq in the conda packages:

https://github.com/ContinuumIO/anaconda-issues/issues/8932#issuecomment-374990875

sigiesec:
@bluca We don't seem to have a corresponding test case for libzmq itself. Would be something like this:
```
void test_push_pull_with_domain ()
{
    void *push = test_context_socket (ZMQ_PUSH);
    const char *zap_domain = "global";
    TEST_ASSERT_SUCCESS_ERRNO (
      zmq_setsockopt (push, ZMQ_ZAP_DOMAIN, zap_domain, strlen (zap_domain)));
    char my_endpoint[MAX_SOCKET_STRING];
    bind_loopback_ipv4 (push, my_endpoint, sizeof (my_endpoint));
    void *pull = test_context_socket (ZMQ_PULL);
    TEST_ASSERT_SUCCESS_ERRNO (zmq_connect (pull, my_endpoint));

    const char *test_string = "test";
    send_string_expect_success (push, test_string, 0);
    recv_string_expect_success (pull, test_string, 0);

    test_context_socket_close (pull);
    test_context_socket_close (push);
}
```
and this fails, too.

I think you implemented only a special handling of zap_enforce_domain for PLAIN, but not for NULL.

bluca:
I knew I'd forget something...

@minrk can you monkey-patch the bundled version? Or amend that test

bluca:
meanwhile I'll work on libzmq as soon as I have time, hopefully today

minrk:
I don't think I should ship a patched libzmq bundled in pyzmq. I'll wait for a fixed libzmq release before bundling libzmq 4.2.

bluca:
Not even a backport of a merged patch? I've just done a release, and added new stable APIs - to do a new bugfix-only I would need to revert them, and then put them back again...

minrk:
I'd rather not. I don't believe libzmq-4.2.4 bundled with pyzmq should behave differently than pyzmq linked against 'real' libzmq-4.2.4 with the same build options. If it were a patch to the build process that would be okay, but not behavior.

You don't need to revert changes to make a backport release - you can create a branch for 4.2.x, e.g.

```
git checkout -b 4.2.x v4.2.4
git cherry-pick e4b3bd8 1c5a63e
```

I'm also okay waiting for 4.3.x. I don't feel any urgency about upgrading the bundled libzmq since users can always link pyzmq to system or self-installed libzmq.

bluca:
We don't do release branches as per C4 - I've pushed 4.2.5, I've added the same test so hopefully it's fine now: https://github.com/zeromq/libzmq/releases/tag/v4.2.5

bluca:
Just gave it a quick spin, the tests that were broken now pass, so if you bump to 4.2.5 it should be good to go.

minrk:
@bluca 4.2.5 works, thanks!

tcwalther:
The PyZMQ documentation on readthedocs now states that 17.0.0 ships with ZeroMQ 4.2.5, although it actually ships with 4.1.6. I was just tripping over this when I tried to use the new constant `zmq.BLOCKY` (introduced in ZeroMQ 4.2). Would it be possible to publish a version 17.0.1 with the updated ZeroMQ?

PR #1161, opened 2018-04-06
Adding stdc++ into libzmq libraries under PyPy is not required

### LINKED ISSUE #1160: It's not required to force to link to stdc++ in libzmq extension on PyPy on Linux anymore
Change including ``stdc++`` into libraries for ``libzmq`` extension for PyPy made in commit https://github.com/zeromq/pyzmq/commit/8d89d4c3604944cd95bb260760a25fb0112d6358 seems to be not required anymore. I've disabled it in ``setup.py``, built ``pyzmq`` and indeed ``libzmq.pypy-41.so`` is linked against ``libstdc++.so.6``. I used PyPy-5.10.0 and setuptools-39.0.1 and PyPy-5.9.0 and setuptools-28.8.0 with gcc-4.6.3, gcc-4.8.2 and gcc-6.4.0 to check my assumption.

The main reason why this is an issue for me is that i'd like to link ``libstdc++`` statically using ``LDFLAGS`` to be able to build zmq with newer compiler and use that module on systems with older ``libstdc++``.

If there will be no objections against removing that ``stdc++`` inclusion to libraries I'd like to make a pull request.

daa:
More details in https://github.com/zeromq/pyzmq/issues/1160

closes #1160

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -603,11 +603,6 @@ class Configure(build_ext):
                 else:
                     info("ok")
 
-                if pypy:
-                    # seem to need explicit libstdc++ on linux + pypy
-                    # not sure why
-                    libzmq.libraries.append("stdc++")
-
         # copy the header files to the source tree.
         bundledincludedir = pjoin('zmq', 'include')
         if not os.path.exists(bundledincludedir):


minrk:
Build passed, thanks!

daa:
Thank you very much.

PR #1175, opened 2018-05-07
Support for building the extension with statically linked libzmq

cowo78:
Support for building the extension with statically linked libzmq.
Added missing extension dependencies on win32 (libzmq dependencies,
must be explicitly set for static builds).
[TODO] libzmq version autodetection does not work with static libzmq

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -93,7 +93,7 @@ for idx, arg in enumerate(list(sys.argv)):
         configure_idx = idx
     elif arg == 'fetch_libzmq':
         fetch_idx = idx
-    
+
     if arg.startswith('--zmq='):
         sys.argv.pop(idx)
         if configure_idx < 0:
@@ -187,7 +187,7 @@ def check_pkgconfig():
 
 def _add_rpath(settings, path):
     """Add rpath to settings
-    
+
     Implemented here because distutils runtime_library_dirs doesn't do anything on darwin
     """
     if sys.platform == 'darwin':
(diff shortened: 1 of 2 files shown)


minrk:
Makes sense, but CI builds are failing. Maybe some checks for whether static builds should be done are incorrect?

setup.py
```diff
@@ -363,13 +367,18 @@ def init_settings_from_config(self):
         )]
         if sys.platform.startswith('win') and sys.version_info < (3, 3):
             settings['include_dirs'].insert(0, pjoin('buildutils', 'include_win32'))
-        
+
+        settings.setdefault('libraries', [])
+        # Explicitly link dependencies, not necessary if zmq is dynamic
+        if sys.platform.startswith('win'):
+            settings['libraries'].extend(('ws2_32', 'iphlpapi', 'advapi32'))
```
minrk: If this is only needed when libzmq is static, is it intentional that the linking unconditional?

cowo78: I just prefer it to be explicit. Easy to change, though.

minrk: ok, thanks

setup.py
```diff
@@ -603,6 +612,11 @@ def bundle_libzmq_extension(self):
                 else:
                     info("ok")
 
+                if pypy:
+                    # seem to need explicit libstdc++ on linux + pypy
+                    # not sure why
+                    libzmq.libraries.append("stdc++")
```
daa: Excuse me for disturbing you but in #1161 I intentionally disabled adding ``libstdc++`` to explicitly linked libraries under PyPy. Do you have any reasons to enable it again? And if you do please tell me, may be I'm missing something.

cowo78: Uhmmm no. Likely my bad while merging.

PR #1183, opened 2018-06-25
default ioloop to current instead of instance

minrk:
since this is tornado's own behavior.

Should result in no difference when using only one thread, but correct behavior when using more than one.


confiq:
I was trying to fix unittests but since this merge with commit 0c44bdb the unittests started to fail

PR #1184, opened 2018-06-26
Asyncio example for pub/sub with dealer/router

NumesSanguis:
Example of how to influence the messages a publisher sends by making use of a router in the same class. Demonstrating how pub-sub-dealer-router sockets can all operate in a single Context in a single class asynchronously.

Tested with:

- Python 3.6
- libzmq version is 4.2.5
- pyzmq version is 17.0.0


minrk:
thanks!

PR #1186, opened 2018-07-03
Fix cleaning cython sources.

eli-schwartz:
Arch Linux is currently rebuilding the world for python 3.7, but pyzmq has outdated cython files. This shouldn't be a problem except the option to rebuild them is completely broken...


eli-schwartz:
Although currently even after backporting #1132 the testsuite is hanging on
```
zmq/tests/asyncio/test_asyncio.py::TestAsyncIOSocket::test_recv <- zmq/tests/asyncio/_test_asyncio.py
```

But I've got no idea why. :/

minrk:
Thanks!

PR #1201, opened 2018-07-23
Remove very outdated Cython compatibility measures

jdemeyer:
1. Remove imports of `bytes`, `unicode` and `basestring`. Cython knows about these classes, regardless of Python version. This is actually an optimization, since Cython optimizes those as builtins but not when imported.

2. Remove typedefs for adding `const`. Cython has supported that for a long time now, there is no need anymore for such hacks.

3. Remove a few redundant assignments (only in code that I was changing anyway). Generally, there is no need for an explicit assignment
```
cdef const char *buf = python_bytes_object
```
to convert bytes to `const char *`.

### DIFF
--- a/zmq/backend/cython/error.pyx
+++ b/zmq/backend/cython/error.pyx
@@ -23,23 +23,17 @@
 # Imports
 #-----------------------------------------------------------------------------
 
-# allow const char*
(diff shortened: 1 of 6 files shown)


minrk:
Thanks!

PR #1215, opened 2018-08-19
optimize async sends

### LINKED ISSUE #1209: zmq.PUB in zmq.asyncio is 7x slower than plain zmq
I was trying to rewrite my PubSub model with zmq.asyncio but it seems really slow vs plain zmq. 
Any thoughts?

This run was on Windows 10, Python 3.6.6, pyzmq 17.1.0 .
Also tried on AWS as Ubuntu 16.04, Python 3.6.5, pyzmq 17.0.0, relatively similar results

**Straight zmq:**
That took: 514.5178ms
Whole thing took: 568.4677ms

**zmq.asyncio:**
That took: 3788.8812ms
Whole thing took3936.3975ms

```python
# zmq_server.py
import zmq
from timeit import default_timer as timer
# import ujson as json
import json # took out the ujson as not everyone had it installed  but it's alot faster

def pub(num):
  port = "5556"
  context = zmq.Context()
  socket = context.socket(zmq.PUB)
  socket.bind("tcp://*:%s" % port)

  myList = ["AAA","BBB","CCC"]
  market = dict()
  market[myList[0]]=[100,10.00,10.01,200]
  market[myList[1]]=[200,10.00,10.01,400]
  market[myList[2]]=[300,10.00,10.01,600]

  start = timer()
  for i in range(num):
    channel = myList[i%3]
    messagedata = market[channel]
    jsoned_data = json.dumps(messagedata)
    socket.send_string(f"{channel} {jsoned_data}")
  print(f"That took: {(timer()-start)*1000}ms")

num = 100000
start = timer()
pub(num)
print(f"Whole thing took: {(timer()-start)*1000}ms")
```

```python
# zmq_asyncio_server.py
import asyncio
import zmq.asyncio
from zmq.asyncio import Context
from timeit import default_timer as timer
# import ujson as json
import json # took out the ujson as not everyone had it installed but it's alot faster

async def pub(num):
  port = "5556"
  context = Context.instance()
  socket = context.socket(zmq.PUB)
  socket.bind("tcp://*:%s" % port)

  myList = ["AAA","BBB","CCC"]
  market = dict()
  market[myList[0]]=[100,10.00,10.01,200]
  market[myList[1]]=[200,10.00,10.01,400]
  market[myList[2]]=[300,10.00,10.01,600]

  start = timer()
  for i in range(num):
    channel = myList[i%3]
    messagedata = market[channel]
    jsoned_data = json.dumps(messagedata)
    await socket.send_string(f"{channel} {jsoned_data}")

  print(f"That took: {(timer()-start)*1000}ms")

num = 100000
producer_coro = pub(num)

start = timer()
loop.run_until_complete(producer_coro)
print(f"Whole thing took{(timer()-start)*1000}ms")
loop.close()
```

minrk:
- small optimizations of attribute-access in zmq.sugar
- use lower-level `.get` calls instead of attribute-access convenience APIs in internal code
- reduce unnecessary checks of zmq.EVENTS
- optimize optimistic async sends

The main change here is in the behavior of async sends that would resolve immediately where DONTWAIT is not given. If no sends are already waiting, a send is attempted immediately, with DONTWAIT (this was already the behavior if the caller passed DONTWAIT explicitly). Only if that send fails does the full async machinery kick in.

The benefit comes from skipping the check for zmq.EVENTS prior to send. Actually blocking sends will take a little longer, since they will attempt and fail to send (with EAGAIN), before being scheduled on the eventloop when the socket is ready.

closes #1209

### DIFF
--- a/zmq/_future.py
+++ b/zmq/_future.py
@@ -65,7 +65,7 @@ class _AsyncPoller(_zmq.Poller):
                 if mask & _zmq.POLLOUT:
                     evt |= self._WRITE
                 self._watch_raw_socket(loop, socket, evt, wake_raw)
-        
+
         def on_poll_ready(f):
             if future.done():
                 return
@@ -176,30 +176,35 @@ class _AsyncSocket(_zmq.Socket):
     
     def recv(self, flags=0, copy=True, track=False):
         """Receive a single zmq frame.
-        
+
         Returns a Future, whose result will be the received frame.
-        
+
         Recommend using recv_multipart instead.
         """
         return self._add_recv_event('recv',
             dict(flags=flags, copy=copy, track=track)
         )
-    
+
     def send_multipart(self, msg, flags=0, copy=True, track=False, **kwargs):
         """Send a complete multipart zmq message.
-        
+
         Returns a Future that resolves when sending is complete.
         """
-        kwargs.update(dict(flags=flags, copy=copy, track=track))
+        kwargs['flags'] = flags
+        kwargs['copy'] = copy
+        kwargs['track'] = track
         return self._add_send_event('send_multipart', msg=msg, kwargs=kwargs)
-    
+
     def send(self, msg, flags=0, copy=True, track=False, **kwargs):
         """Send a single zmq frame.
-        
+
         Returns a Future that resolves when sending is complete.
-        
+
         Recommend using send_multipart instead.
         """
+        kwargs['flags'] = flags
+        kwargs['copy'] = copy
+        kwargs['track'] = track
         kwargs.update(dict(flags=flags, copy=copy, track=track))
         return self._add_send_event('send', msg=msg, kwargs=kwargs)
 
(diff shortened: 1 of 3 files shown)


minrk:
Results from the benchmark in #1209:

17.1.2:

- zmq_server: 754ms
- zmq_asyncio_server: 7103ms (9.4x)
- zmq_asyncio_server optimized: 1049ms (1.4x)

    ```python
    f = socket.send(f"{channel} {jsoned_data}".encode("utf8"), zmq.DONTWAIT)
    if not f.done():
      await f
    ```

this PR:

- zmq_server: 1x
- zmq_asyncio_server: 1206ms (1.6x)
- zmq_asyncio_server optimized (without DONTWAIT): 1049ms (1.4x)

The main change is that sends that would be resolved immediately (the vast majority of the time) take the same optimized path as when DONTWAIT is given in 17.1.2.

PR #1230, opened 2018-09-26
Fixed Pypi link

SayakMukhopadhyay:
Link was wrongly pointed to https://pypi.org/projects/pyzmq/ instead of `project`


minrk:
Thanks!

PR #1252, opened 2019-01-14
Problem: zmq.EVENT_ALL changes as more events are added and test fails

bluca:
Solution: pin the test so that it only requests the events it's looking
for, to avoid failures when a new event is added.
As the documentation says, and as it can be expected, EVENT_ALL is a
catch-all mask that changes over time.


bluca:
@minrk in libzmq 4.3.0 a bunch of new events were moved to stable, so the _ALL mask now catches them too

minrk:
Thanks!

PR #1258, opened 2019-02-08
update bundled libzmq to 4.3.1

### LINKED ISSUE #1145: Unknown event 16384 on monitoring socket
I'm using the following code to map event numbers to names:

    EVENT_MAP = {}
    for name in dir(zmq):
        if name.startswith("EVENT_"):
            value = getattr(zmq, name)
            EVENT_MAP[value] = name
    ...
    event = zmq_monitor.recv_monitor_message(self._monitor, flags=zmq.NOBLOCK)
    event.update({"description": self.EVENT_MAP[event["event"]]})
    logger.debug(__("Event: {0}", event))

After updating zeromq and pyzqm yesterday this now fails with:

    KeyError: 16384

Event 16384 (0x4000) is defined in src/zmq_draft.h as

    #define ZMQ_EVENT_HANDSHAKE_FAILED_AUTH        0x4000

I have compiled neither zeromq nor pyzmq with any --enable-drafts option so I'm surprised that I would get a draft event from a monitoring socket. If zmq enables drafts per default then pyzmq should do so too.

### LINKED ISSUE #1256: Is pyzmq affected with Remote code execution vulnerability (CVE-2019-6250)
We could see the libzmq has affected with the vulnerability (CVE-2019-6250) and fix also released(CVE-2019-6250). So would like to know whether the pyzmq is also affected with the same. If yes, then can we have the patch to be released.

minrk:
- bundled libzmq is 4.3.1
- update constants, sockopts with changes
- <del>define zmq.PYZMQ_EVENT_ALL mask that includes all known events and use this as the default for monitor methods instead of zmq.EVENT_ALL</del>

closes #1256
closes #1145


zmq/backend/cython/socket.pyx
```diff
@@ -633,44 +633,53 @@ cdef class Socket:
...
-        
+
         .. versionadded: libzmq-3.2
         .. versionadded: 14.0
-        
+
+        .. versionchanged: 18.0
+            default to new zmq.PYZMQ_EVENT_ALL,
+            which is all events known by pyzmq,
+            instead of zmq.EVENT_ALL,
```
bluca: note that we are changing this, and making the EVENT_ALL fixed - the monitor stuff is now versioned independently, and if you don't ask for the new version you won't get new events in the future

minrk: Thanks. I went ahead and reverted this, leaving it up to libzmq for how to behave.

PR #1275, opened 2019-03-26
Teardown timeout are too short on slow systems

Carreau:
Like for exampel Travis-CI. Make them configurable with a class
attribute. This shoudl make it easy as an escape hatch for subproject to
change the timeout and have more reliable test without having to
re-write the whole tearDown just to change one parameter.


minrk:
Makes sense, thanks!

Issue #1286, opened 2019-05-02
Compilation error with Python 3.8a3


zopyx:
```
    gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -Izmq/utils -Izmq/backend/cython -Izmq/devices -c build/temp.linux-x86_64-3.8/scratch/vers.c -o build/temp.linux-x86_64-3.8/scratch/vers.o
    gcc -pthread build/temp.linux-x86_64-3.8/scratch/vers.o -lzmq -lrt -o build/temp.linux-x86_64-3.8/scratch/vers
        ZMQ version detected: 4.1.6
    Warning: Detected ZMQ version: 4.1.6, but pyzmq targets ZMQ 4.3.1.
    Warning: libzmq features and fixes introduced after 4.1.6 will be unavailable.
    ************************************************
    ************************************************
    building 'zmq.backend.cython._device' extension
    creating build/temp.linux-x86_64-3.8/zmq
    creating build/temp.linux-x86_64-3.8/zmq/backend
    creating build/temp.linux-x86_64-3.8/zmq/backend/cython
    gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -DHAVE_SYS_UN_H=1 -Izmq/utils -Izmq/backend/cython -Izmq/devices -I/home/ajung/src/pp.server/include -I/home/ajung/.pyenv/versions/3.8-dev/include/python3.8 -c zmq/backend/cython/_device.c -o build/temp.linux-x86_64-3.8/zmq/backend/cython/_device.o
    zmq/backend/cython/_device.c: In function ‘__Pyx_InitCachedConstants’:
    zmq/backend/cython/_device.c:2670:89: warning: passing argument 6 of ‘PyCode_New’ makes integer from pointer without a cast [-Wint-conversion]
[...]

PR #1292, opened 2019-05-30
BLD: pin cython language version to 2

tacaswell:
This is the language that level that cython expects in the pyx and pyd
files not the target.  Most critically, the implicit local imports are
not supported by the new default ('3str').

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1146,7 +1146,7 @@ ext_kwargs = {
 }
 if cython:
     # set binding so that compiled methods can be inspected
-    ext_kwargs['cython_directives'] = {'binding': True}
+    ext_kwargs['cython_directives'] = {'binding': True, 'language_level': 2}
 
 for submod, packages in submodules.items():
     for pkg in sorted(packages):


tacaswell:
All of the travis failures look like they are 

```
E: Failed to fetch http://dl.google.com/linux/chrome/deb/dists/stable/main/binary-amd64/Packages.gz  Hash Sum mismatch
E: Some index files failed to download. They have been ignored, or old ones used instead.
```

power-cycling to retrigger CI

-----

I need this patch to compile pyzmq with cython and cpython master branches.  This change is required due to a change in cython, but you need very new cython to support py38.  If people want I can track down exactly what versions this is required for.

tacaswell:
Related to https://github.com/zeromq/pyzmq/issues/1286

tacaswell:
The failures are py27 build that hang, also see that on the last master build https://travis-ci.org/zeromq/pyzmq/builds/511379309

minrk:
Hangs appear related to gevent. Maybe an update over there, and probably not related to this PR.

Thanks!

I'd like to move to language_version: 3 soon as well.

PR #1300, opened 2019-06-25
Add Python 3 code in NumPy serialization example

MasonMcV:
In Python 3.x, the buffer builtin is no longer used.  Instead this is called memoryview.
Updated the example code to include both of these, with a note explaining which to use.


minrk:
Indeed, and `memoryview` is in Python 2.7, so the python 3 way will work everywhere. Plus, Python 2 is itself end-of-life this year, so the examples should all be Python 3 syntax by default anyhow. Thanks!

PR #1301, opened 2019-06-26
Display ipc path in case of ENOENT exception

Peque:
Useful to see right in the traceback which ipc path caused the
exception.

### DIFF
--- a/zmq/tests/test_socket.py
+++ b/zmq/tests/test_socket.py
@@ -3,6 +3,7 @@
 # Distributed under the terms of the Modified BSD License.
 
 import copy
+import errno
 import json
 import os
 import platform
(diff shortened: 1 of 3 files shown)


zmq/tests/test_socket.py
```diff
@@ -396,7 +397,19 @@ def test_ipc_path_max_length_msg(self):
...
+    def test_ipc_path_no_such_file_or_directory_message(self):
+        """Display the ipc path in case of an ENOENT exception"""
+        s = self.context.socket(zmq.PUB)
+        self.sockets.append(s)
+        invalid_path = '/foo/bar'
+        with pytest.raises(zmq.ZMQError) as error:
+            s.bind('ipc://{0}'.format(invalid_path))
+        error_message = str(error.value)
+        assert invalid_path in error_message
+        assert "no such file or directory" in error_message.lower()
```
minrk: Can you add one more assert that `error.errno == ENOENT`?

Peque: Thanks for the review!

I pushed a fixup with that change.

Tell me if it is good for you and I will squash it. Or feel free to squash it yourself before merging. :+1:

minrk: no need for that. Thanks!

PR #1306, opened 2019-06-29
Propagate cancellation in AsyncSocket.poll when socket closes

### LINKED ISSUE #1305: Closing an async socket raises error in poll method
When directly polling a socket with asyncio, an uncatchable CancellationError is raised, similiar to what happened in #1114 

Version info
```
libzmq-4.3.1
pyzmq-18.0.2
Python-3.7.3 (default, Apr  3 2019, 19:16:38) 
[GCC 8.0.1 20180414 (experimental) [trunk revision 259383]]
```
some code to reproduce
```
import asyncio
import zmq
from zmq.asyncio import Context


async def start_polling(socket):
    try:
        print('starting')
        await socket.poll()
    except asyncio.CancelledError:
        print('Got cancelled')
    finally:
        print('done')


async def stop(socket):
    await asyncio.sleep(1)
    socket.close()
    print('socket closed')


async def main():
    socket = Context.instance().socket(zmq.PULL)
    await asyncio.gather(
        start_polling(socket),
        stop(socket)
    )

asyncio.run(main(), debug=True)
```

elfjes:
fixes #1305

### DIFF
--- a/zmq/_future.py
+++ b/zmq/_future.py
@@ -254,6 +254,13 @@ class _AsyncSocket(_zmq.Socket):
         def unwrap_result(f):
             if future.done():
                 return
+            if f.cancelled():
+                try:
+                    future.cancel()
+                except RuntimeError:
+                    # RuntimeError may be called during teardown
+                    pass
+                return
             if f.exception():
                 future.set_exception(f.exception())
             else:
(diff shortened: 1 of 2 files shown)


minrk:
Excellent! Would you be able to add a test for this?

elfjes:
Testing it is a bit wonky. I would do something like:

```python
a, b = self.create_bound_pair(zmq.PUSH, zmq.PULL)

f = b.poll(timeout=1)
b.close()
with pytest.raises(asyncio.CancelledError):
    yield from f
```
but this completely stalls the test before fixing the bug. So this is the best I could do...

minrk:
That works, thanks!

Issue #1315, opened 2019-08-01
fails to install with Python 3.8.0b3


cs01:
I stumbled on this while testing pipx for python 3.8 compatibility and wanted to report it here. 

```
python --version
Python 3.8.0b3
```

```
pip install pyzmq
...
    In file included from /home/csmith/Downloads/Python-3.8.0b3/Include/compile.h:5:0,
                     from /home/csmith/Downloads/Python-3.8.0b3/Include/Python.h:138,
                     from zmq/backend/cython/_device.c:4:
    /home/csmith/Downloads/Python-3.8.0b3/Include/code.h:122:28: note: declared here
     PyAPI_FUNC(PyCodeObject *) PyCode_New(
                                ^~~~~~~~~~
    error: command 'gcc' failed with exit status 1
    ----------------------------------------
ERROR: Command errored out with exit status 1: /tmp/pvenv/bin/python -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-6mnvkxwv/pyzmq/setup.py'"'"'; __file__='"'"'/tmp/pip-install-6mnvkxwv/pyzmq/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /tmp/pip-record-57y2pv8f/install-record.txt --single-version-externally-managed --compile --install-headers /tmp/pvenv/include/site/python3.8/pyzmq Check the logs for full command output.
```

Full log: https://gist.github.com/cs01/174f417c0857e11d9beffa4d17e600ac

bjodah:
I believe the sdist may contain rendered Cython sources. Installing from the repo using latest cython work for me either way as workaround for now.

kinow:
Building from sources with python 3.8.0b3 and latest cython also working for me. Issue only when using pip.

minrk:
Yes, I need to publish another release with latest Cython, then I think it should be okay. I'll try to do that this week.

minrk:
Published 18.1 with latest cython, so this should work on 3.8b3

PR #1316, opened 2019-08-03
Check for asyncio.exceptions.CancelledError when py38+ in unit test

kinow:
Hi,

I was testing an application I work on and that uses `pyzmq` with the latest version of Python. While setting up the environment, I ran into #1315 , then decided to build locally (which works fine). But running the tests with Python 3.8.0b3 (built from master yesterday), gives me a `CancelledError` on this test.

Looks like after 3.8, `asyncio` will raise `asyncio.exceptions.CancelledError`, not `concurrent.futures.CancelledError`, due to https://bugs.python.org/issue32528. So added an `if` similar to another one in the same file, but checking for 3.8.

That way this test passed on Python 3.8.0b3, and also on local Python Anaconda 3.7.3.

Cheers
Bruno


minrk:
Thanks!

PR #1321, opened 2019-08-17
Example project on how to combine Docker with ZeroMQ for micro-services

NumesSanguis:
Docker allows for deploying services on any system without going through the trouble of installing packages or any compiling.
ZeroMQ + Docker are a natural fit for creating micro-services by having multiple Docker containers communicate with each other over ZMQ.
This project contains an example Publisher-Subscriber that can be set-up in a Python-Python, Docker-Docker and Docker-Python fashion, without any code changes.

It is a copy from my repository: https://github.com/NumesSanguis/pyzmq-docker
It is BSD licensed.

Hope it can be useful for fellow ZeroMQ/PyZMQ folks :)


NumesSanguis:
Could someone explain me why AppVeyor failed with this?:
```
start /W pytest -vsx -m "new_console" --pyargs zmq.tests --junit-xml=../results.xml
python ../tools/check_junit_result.py ../results.xml
No tests run
Command exited with code 1
```

examples/docker/sub/main.py
```diff
@@ -0,0 +1,37 @@
...
+import zmq
+
+
+def subscriber(ip="0.0.0.0", port=5551):
+    # ZMQ connection
+    url = "tcp://{}:{}".format(ip, port)
+    print("Going to bind to: {}".format(url))
+    ctx = zmq.Context()
+    socket = ctx.socket(zmq.SUB)
+    socket.bind(url)  # subscriber creates ZeroMQ socket
```
JSalazar88, line 14: If this is a subscriber, shouldn't this be a `socket.connect(url)` call?

examples/docker/pub/main.py
```diff
@@ -0,0 +1,44 @@
...
+import time
+
+
+def publisher(ip="0.0.0.0", port=5551):
+    # ZMQ connection
+    url = "tcp://{}:{}".format(ip, port)
+    print("Going to connect to: {}".format(url))
+    ctx = zmq.Context()
+    socket = ctx.socket(zmq.PUB)
+    socket.connect(url)  # publisher connects to subscriber
```
JSalazar88, line 15: If this is a publisher, shouldn't this be a `socket.bind` call?

NumesSanguis: Thank you for checking @JSalazar88. I agree that it's more common that a single Publisher sends message to multiple Subscribers, but with ZMQ it is also possible to have a single Listener wait for messages from multiple Publishers.
In the example code, there is only 1 Pub and 1 Sub, so it doesn't really matter which socket connects and which one binds. This example is now setup as a m-Pub to 1-Sub pattern. Could be changed, but it won't change anything for the functioning of this example.

examples/docker/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
+version: "3"
+services:
+  sub:
+    build:
+      context: ./sub  # Docker context from folder of this file; needed to include requirement.txt
+      dockerfile: Dockerfile
+    ports:
+      - "5550:5550"  # map container interal 5550 port to publicly accessible 5550 port
```
Lard4, line 8: why 5550 and not 5551 as the defaults would suggest?

NumesSanguis: It has been a while ago I wrote this, but I think I did this so you could both test the Docker and non-Docker version at the same time. If the port value is the same, and you start them both, one of the 2 cannot bind.

PR #1338, opened 2019-11-04
Update Pip compilation instructions

acairncross:
`--no-use-wheel` has been deprecated since Pip v7.0.0, and has stopped working
at some point.

https://pip.pypa.io/en/stable/news/#id242


minrk:
Thanks!

PR #1347, opened 2019-12-04
More useful PUB-SUB logging with pyzmq

aeetos:
I recently discovered log.PUBHandler and was intrigued by the potential for distributed Python logging with zmq pub-sub sockets. When I tried it, I encountered a few issues that made it difficult to use and somewhat un-pythonic. 

I had to wade into the source to learn what was happening, especially when I tried to set my own custom log format, and there was little guidance on how to build my own viewer for the messages.

I believe these changes will address those issues and make the module more attractive, even for developers who are not yet using ZMQ for anything else in their project. It should now be possible to use pyzmq for distributed logging with almost no knowledge of how ZMQ works.

__Highlights__
* Default value for root_topic ensures consistent out-of-the-box behavior - subscriptions to 'root' just work without having to set a class-level property. Choose 'root' as the default value because that is the default name of the root logger in the Python logging module.

* New documented setRootTopic method emphasizes its importance. Property setters/getters for root_topic provide backward compatibility (confirmed by passing unit tests). Blank topics still work (also confirmed by unit tests).

* New documented setFormatter method for overriding log formats is consistent with the rest of the Python logging module. Preserves existing behavior that selectively formats messages by level.

* New module that provides a command-line viewer for published log messages, executable as "python -m zmq.log tcp://127.0.0.1:1234". Includes optional color and alignment switches and selective filtering by topic.

I also had to do a bit of learning to sort out how to get a working dev environment, likely because this is my first time contributing to a project with c libraries running behind it. I made an attempt to document this in CONTRIBUTING, but it bears close review from the experts. 

Finally, there's a few additions to your .gitignore file that work for me but may be better as personal settings instead of project settings. Happy to revert that if you disagree with it being in the project.

I'd be glad to add some documentation for this once I know how much of it you're likely to bring into the main project. 

Hope this helps! Looking forward to seeing feedback on it. Thanks for reviewing it.

-- Mark Kendrick


zmq/log/handlers.py
```diff
@@ -80,9 +81,52 @@ def __init__(self, interface_or_socket, context=None):
...
+        from this handler.
+
+        You may also set the root topic to an empty string. In this case, messages
+        will begin with the binary representation of the log level string (INFO, WARN, etc.).
+        Note that ZMQ SUB sockets can have multiple subscriptions.
+        """
+        self._root_topic = root_topic
+
+    @property
+    def formatters(self):
```
minrk: Is there a reason formatters is a property? The getters and setters don't appear to do anything that a plain formatters attribute wouldn't do.

CONTRIBUTING.md
```diff
@@ -83,4 +83,50 @@ libzmq, and inherits LGPL
...
+```
+python setup.py configure --zmq=/usr/local
+python setup.py build_ext --inplace
+```
+
+
+If all went according to plan you should be able to run the unit tests without a problem. For example:
+
+```
+python -m unittest zmq.tests.test_pubsub
```
minrk:
```suggestion
pytest zmq/tests/test_pubsub.py
```

CONTRIBUTING.md
```diff
@@ -83,4 +83,50 @@ libzmq, and inherits LGPL
...
 
 [tornado]: http://www.tornadoweb.org
-[paramiko]: http://www.lag.net/paramiko
\ No newline at end of file
+[paramiko]: http://www.lag.net/paramiko
+
+## Setting up development environment
+
+To start development on pyzmq, you need a working development environment. Due to the dependancies on libzmq,
+this is a bit more involved than simply checking out the code.
```
minrk: In almost any context, `pip install -r test-requirements -e .` will produce a development build. Specifying a non-bundled zeromq with `--zmq=...` is also possible, but not generally necessary.

CONTRIBUTING.md
```diff
@@ -83,4 +83,50 @@ libzmq, and inherits LGPL
...
+
+
+If all went according to plan you should be able to run the unit tests without a problem. For example:
+
+```
+python -m unittest zmq.tests.test_pubsub
+```
+
+
+[PyZMQ wiki]:[https://github.com/zeromq/pyzmq/wiki/Building-and-Installing-PyZMQ]
```
minrk: Oof. I had forgotten there even was a Wiki (as you can see from it not being updated in six years). Let's not link to it.

CONTRIBUTING.md
minrk: It looks like these instructions were taken mostly from the extremely out-of-date wiki. I updated the page with more accurate and simple instructions. This isn't nearly so complicated now.

zmq/log/handlers.py
```diff
@@ -56,22 +56,23 @@ class PUBHandler(logging.Handler):
...
-        logging.INFO: logging.Formatter("%(message)s\n"),
-        logging.WARN: logging.Formatter(
-        "%(levelname)s %(filename)s:%(lineno)d - %(message)s\n"),
-        logging.ERROR: logging.Formatter(
-        "%(levelname)s %(filename)s:%(lineno)d - %(message)s - %(exc_info)s\n"),
-        logging.CRITICAL: logging.Formatter(
-        "%(levelname)s %(filename)s:%(lineno)d - %(message)s\n")}
     
-    def __init__(self, interface_or_socket, context=None):
+    def __init__(self, interface_or_socket, context=None, root_topic='root'):
```
minrk: I'm not sure 'root' makes the most sense as a default root topic. No default makes more sense to me. Could it be that this should be the logger name by default? I'm not sure that's feasible, though, since the Handler doesn't have a handle on the Logger object.

zmq/log/handlers.py
aeetos: Sure. My intent was to move formatters from a class-level attribute to a _protected attribute without breaking any existing code that expected them to be an attribute. Similar to what I did with root_topic to _root_topic. 

Really, I was hoping we could deprecate attribute access entirely and focus on the new setFormatter method instead. Right now, the only way to know how to set that attribute to the correct value is to look at the code inside handlers.py, learn it's a mapping indexed by the log level constant value, and with values of type logging.Formatter. I guess I'm proposing doing away with that direct attribute access. Maybe raise a DeprecationWarning on these properties? Or take the leap that no one is accessing them today, and just remove the properties. 

Either way is fine with me - if you'd prefer I revert formatters back to an attribute and remove these properties, I'll gladly make that happen. LMK.

CONTRIBUTING.md
aeetos: Brilliant! I tried your instructions on the wiki and they work perfectly, with just one small error:

`pip install -r test-requirements`
should be
`pip install -r test-requirements.txt`

Based on that, I'd say everything I added in CONTRIBUTING.md isn't necessary, and if you agree I'll just revert those changes. Or maybe link to the wiki? LMK.

CONTRIBUTING.md
aeetos: No worries - see my comments above about CONTRIBUTING.md - looks like my changes there can go poof.

zmq/log/handlers.py
aeetos: I really like the idea of the topic following the logger name, but as you said that may not be feasible, especially with how PUBHandler is built. It's designed to build up the message topic from the root topic, then the log level, then a subtopic passed in with a :: in the message. Seems tricky to modify that without changing existing behavior.

After we get past this, I was considering adding a new handler class that did use the logger name as the topic, because then the topics could follow the package hierarchy, and you could subscribe to log notifications based on where a module is in the layout of your codebase. How neat would that be!

I'm pretty sure the logger name would be available to do that - it is one of the attributes of LogRecord objects. 

But for this, as it is today, I don't care too much about what we call it, but I do think it shouldn't be blank. 

When I first started with this class, I inadvertently set the attribute `roottopic` instead of the correct `root_topic`. I got no errors (another reason I dislike direct attribute access in object APIs) but also no messages. I finally set my subscription to a blank value to start getting anything. That was confusing especially for a ZMQ newbie like myself - I didn't realize until very recently that you could have a "wildcard" subscription. 

My thinking is that if the value is populated with a default value, client code like the log watcher I built can "just work" out of the box, by subscribing to an expected value. But, we could get to the same outcome if we defaulted to a blank value, and then my log watcher could just subscribe to the empty string. I can happily make that change if you'd prefer - I just wasn't sure if "wildcard" subscriptions were to be encouraged in the ZMQ world.

Oh, and the reason I choose the string 'root' is because that is the name of the root logger in the core Python library. Could easily be something else.

CONTRIBUTING.md
```diff
@@ -83,4 +83,50 @@ libzmq, and inherits LGPL
 * perf examples are (c) iMatix, and LGPL
 
 [tornado]: http://www.tornadoweb.org
-[paramiko]: http://www.lag.net/paramiko
\ No newline at end of file
+[paramiko]: http://www.lag.net/paramiko
+
+## Setting up development environment
+
+To start development on pyzmq, you need a working development environment. Due to the dependancies on libzmq,
```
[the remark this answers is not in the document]
aeetos: Good catch, thanks! Sounds like I might be reverting my changes to this file in favor of the wiki, but if that's not how this goes I'll be sure to fix this.

aeetos:
Hi everyone - just a gentle post-holiday bump on this. If we think this has merit, I'd like to get your concerns addressed before we diverge too far from master.  I'll gladly submit a new commit once I get your response on my comments. Thanks!

CONTRIBUTING.md
minrk: Yeah, feel free to revert these for now.

zmq/log/handlers.py
minrk: Since this is a Handler and we can't feasibly match the logger name, let's leave the existing behavior of no default root topic. With tree-based subscriptions in zmq PUB-SUB, the 'root' is really the empty string, not the word 'root'. Lacking a topic really is a valid and reasonable default behavior, and I'd personally like to leave that.

> I just wasn't sure if "wildcard" subscriptions were to be encouraged in the ZMQ world.

Yeah, it's quite common and in my experience what is done almost all of the time.

zmq/log/handlers.py
minrk: The nice thing about properties and attributes is that there's no distinction for API consumers. Making them attributes if there are no setter actions works, then if we discover a need for a setter action, making it a property later is completely transparent to the user. Let's leave formatters as an attribute.

aeetos: Gladly - should be done now.

CONTRIBUTING.md
aeetos: I reverted changes to CONTRIBUTING so this doesn't apply now.

CONTRIBUTING.md
aeetos: I reverted changes to CONTRIBUTING so this doesn't apply now.

zmq/log/handlers.py
aeetos: Okay, done! Let me know if you see anything I missed.

CONTRIBUTING.md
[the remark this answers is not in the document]
aeetos: I reverted changes to CONTRIBUTING so this doesn't apply now, but thanks again for catching it.

aeetos:
Glad to hear I'm on the right track - no worries on the delay @minrk (and congrats!). 

I added two commits to address the earlier requests, and then made a first attempt at some documentation.

minrk:
Aces! Thanks for your patience.

PR #1357, opened 2020-02-11
Make contributing.md code example py3 compatible

kinow:
Trivial change, while reading the docs noticed it was py2 only. This should work on both I think. Tested locally with Python 3.7.6 :+1:


minrk:
Thanks!

PR #1358, opened 2020-02-12
Problem: build with cython < 0.29 fails

bluca:
Solution: add backward compatibility for language_level and set it to 2
if cython doesn't support 3str

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1126,6 +1126,7 @@ submodules = {
 }
 
 min_cython_version = '0.20'
+cython_language_level = '3str'
 try:
     import Cython
     if V(Cython.__version__) < V(min_cython_version):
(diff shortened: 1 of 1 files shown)


bluca:
@minrk ping

minrk:
For better consistency of installation, can we make the minimum Cython 0.29 instead?

bluca:
There are LTS platforms supported for years to come with older versions, which is how I found out - Debian 8, Debian 9, CentOS 7, Fedora 29, SUSE 42, Ubuntu 16.04, Ubuntu 18.04. Upping the requirement would mean cutting them all off.

minrk:
Those platforms wouldn't expect new feature releases of pyzmq, would they? Shouldn't their package lists be frozen long ago?

Note that pyzmq releases include Cython-generated sources, so installing a release of pyzmq on all those platforms will work fine without Cython at all. Cython is only a development requirement.

bluca:
Up to you - given those are LTSes I thought users could be trying to build from sources. I'll disable the builds we do on OBS for those releases.

minrk:
Updated to require 0.29 on Python >= 3.7 and set 3str if possible, leaving the fallback on Cython's default behavior for old versions of Cython, which is what we were doing before.

Building from official releases doesn't require Cython at all, so I'm not worried about folks having issues installing pyzmq on LTS.

bluca:
Thanks!

Issue #1361, opened 2020-02-17
asyncio: memory leak when registering more than one socket to poller


RoPP:
#968 fixed memory leak for one socket but polling multiple sockets still leak memory

```python
import zmq
import zmq.asyncio
import asyncio

sock1 = zmq.asyncio.Context.instance().socket(zmq.REP)
sock1.connect("tcp://localhost:55570")
sock2 = zmq.asyncio.Context.instance().socket(zmq.REP)
sock2.connect("tcp://localhost:55560")

poller = zmq.asyncio.Poller()
poller.register(sock1, zmq.POLLIN)
poller.register(sock2, zmq.POLLIN)

async def main():
    while True:
        try:
            await poller.poll(timeout=1)
            pass
        except KeyboardInterrupt:
            break
asyncio.run(main())
```
It seems that `_recv_futures` is still expanding indefinitely. The watcher is removed from `_recv_futures` on one socket but remains in others sockets' list.

minrk:
closed by #1362

PR #1362, opened 2020-02-17
fix: avoid memory leak in case of several sockets registered to a poller

Paul-AUB:
Here a fix to manage several sockets (#1361), based on the fact each socket has its own list of Futures.
The fix no more use a flag on the shared instance of Future.

### DIFF
--- a/zmq/_future.py
+++ b/zmq/_future.py
@@ -114,6 +114,9 @@ class _AsyncPoller(_zmq.Poller):
 
 class _AsyncSocket(_zmq.Socket):
 
+
+    # Warning : these class variables are only here to allow to call super().__setattr__.
+    # They be overridden at instance initialization and not shared in the whole class
     _recv_futures = None
     _send_futures = None
     _state = 0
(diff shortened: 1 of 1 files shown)


minrk:
Thanks! I'll need to refresh my memory on what problem that flag solved to see if it's really obsolete and can be removed without being replaced with something else.

minrk:
On rereading #969 which added this, I think this PR is right, thanks!

More detail:

- `_pyzmq_popped` flag was an optimization signal to avoid iterating through a list when the Future is known to have been removed from that list
- `_pyzmq_popped` was set on the Future on removal from the e.g. `send_futures`  list
- the `watcher` Future was in multiple event lists (could be send and recv on one socket, or multiple sockets), so it would only ever get removed from *one* event list, the first time this flag was set, hence the leak.

Removing this check eliminates the optimization - when an event is removed from the event list, when it is done, all Futures in the event list will always be checked, even knowing that it won't match. However, the optimization is likely not worth preserving, given that it needs to be tracked per event list for the poller, and each event list is likely to be at most 1-3 Futures at any given time, so very little was saved.

rocotrade:
This solves the exact problem I've observed with version 18.0.2 - I have 2 sockets registered to an async Poller and it was a memory leak in the library.  
Possibly stupid question but when a bug is fixed and merged into master, does this mean it's just in the current version at the time of the merge, and future versions going forward?  If it was a feature you would actually introduce a new subversion, but never for a bug fix?

minrk:
This patch is new in 19.0.

>  I can't see which is the version this got added to

It's not always the easiest, but through the GitHub UI, clicking on [one of the commits in a PR](https://github.com/zeromq/pyzmq/pull/1362/commits/799b5eafd36527a741b1495642904ce1ca3b46e0) shows you what tags a commit can be found in:

Screen Shot 2020-03-12 at 09 12 45


> looks like the documentation says current version is 18.1.0

Something went wrong on readthedocs and it had stopped auto-building. Thanks for noticing! It should be fixed now. If that hadn't happened, you would have seen the memory leak fix [in the changelog for 19.0](https://pyzmq.readthedocs.io/en/v19.0.0/changelog.html#id1).

PR #1367, opened 2020-02-26
be more cautious about weakref garbage collection

### LINKED ISSUE #1356: Exception ignored in: ... __del__ - close - _rm_socket
After running pytest see this message repeating twice:

Exception ignored in: <function Socket.__del__ at 0x7f83905c59e0>
  File "/Users/wg/miniconda3/envs/abv/lib/python3.7/site-packages/zmq/sugar/socket.py", line 67, in __del__
  File "/Users/wg/miniconda3/envs/abc/lib/python3.7/site-packages/zmq/sugar/socket.py", line 105, in close
  File "/Users/wg/miniconda3/envs/abc/lib/python3.7/site-packages/zmq/sugar/context.py", line 153, in _rm_socket
TypeError: 'NoneType' object is not callable

All test passed and pytest returns 0. (macos, python 3.7.6)

Seem not hurt anything, but would be nice to why? GC collects something more then once? Any way to suppress?

minrk:
weakref.ref may be None during process teardown, don't try to call it if so

closes #1356

### DIFF
--- a/zmq/sugar/context.py
+++ b/zmq/sugar/context.py
@@ -9,6 +9,10 @@ import os
 from threading import Lock
 import weakref
 
+# direct reference limits garbage collection issues
+# during process teardown
+weak_ref = weakref.ref
+
 from zmq.backend import Context as ContextBase
 from . import constants
 from .attrsettr import AttributeSetter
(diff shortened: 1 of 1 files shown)


lvxejay:
When do you expect we'll have a new release with this included?

minrk:
Released 19.0.1 with this fixed

PR #1383, opened 2020-05-22
Avoid unnecessarily importing ctypes

zooba:
This only really impacts Windows, as I believe most other platforms are going to have `getdlopenflags` and follow the same path as before, but it allows us to use `pyzmq` on builds that don't include ctypes. 

I _think_ the rest of the logic for those flags can be simplified, but not significantly, and presumably it's working fine.

FWIW, the actual patch I'm using (given I have a known platform) is just `import .libzmq as _libzmq` instead of the entire function. But if this much is upstream then I don't need to patch pyzmq at all ;)

### DIFF
--- a/zmq/__init__.py
+++ b/zmq/__init__.py
@@ -6,13 +6,14 @@
 # load bundled libzmq, if there is one:
 def _load_libzmq():
     """load bundled libzmq if there is one"""
-    import sys, ctypes, platform, os
+    import sys, platform, os
     dlopen = hasattr(sys, 'getdlopenflags') # unix-only
     # RTLD flags are added to os in Python 3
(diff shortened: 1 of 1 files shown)


minrk:
OK! How do you get Python without ctypes?

zooba:
@minrk Create your own distribution and don't include `_ctypes.pyd` ;)

And I know that seems unreasonable for a widespread public distribution, but for within a business, and particularly for production deployments, it's pretty trivial. Most large companies aren't using a publicly available build of Python already, but haven't quite migrated to private builds/forks of all their dependencies. (And you just saved me from having to maintain a pyzmq fork :) Thanks!)

Issue #1167, opened 2018-04-13
pyzmq 17 asyncio hangs on exit


antonymayi:
after upgrading *just* pyzmq (same libzmq 4.2.5 and python 3.6.5 used in both cases) from 16.0.3 to 17.0 the following simple code runs to the end and then hangs forever while it was exiting ok before:

```
import asyncio
import zmq.asyncio

ctx = zmq.asyncio.Context()
pub = ctx.socket(zmq.PUB)
pub.bind('inproc://test')
sub = ctx.socket(zmq.SUB)
sub.connect('inproc://test')
sub.setsockopt(zmq.SUBSCRIBE, b'')

msg = b'foo', b'bar'

async def run():
    await pub.send_multipart(msg)
    assert msg == tuple(await sub.recv_multipart())
    print('DONE')

asyncio.get_event_loop().run_until_complete(run())
```

produces:
```
$ python xxx.py
DONE
```
and hangs forever.

couldn't find any notes suggesting I would need to change something when upgrading to 17.0 - so is this a bug?

jvansanten:
A random comment from the void: I can reproduce this. It appears that the sockets must be explicitly closed, otherwise their destructors hang forever in zmq/backend/cython/checkrc.pxd.

minrk:
This is a bug. An elusive one. The root cause of this is that a context has its `.term()` called during `__del__` before one or more of its sockets is closed. This will hang until the context is closed.

Closing sockets and contexts via garbage collection during process teardown has been a tricky issue since Python 3.4 changed how that works (in particular, it deliberately increased the randomness of order of the cleanup process).

I was able to reproduce this, and it turned out the `__dealloc__` is being called (which doesn't call `.close()`), but `__del__` is not (which does call close). So the fix should be #1173, which ensures that close is called in `__dealloc__` if it hasn't already been called by `__del__`.

The simplest workaround in the meantime is to call `ctx.destroy()` at the end of your script, which closes all sockets in order and terminates the context.

kalzoo:
Still seeing this with `pyzmq==19.0.1`. Closing the socket (or destroying the context) fixes the problem, but it will be tricky to ensure that always happens. Here's the repro, FWIW. The "bug" is the hang at the end of script.

```
import asyncio
import logging
import zmq
import zmq.asyncio

# Yields a minimal amount of debug information
logging.basicConfig(level=logging.DEBUG)

# Set up our zmq socket
context = zmq.asyncio.Context()
[...]

PR #1390, opened 2020-06-02
#1167 Release reference to sockets on process teardown

sakurai-youhei:
This is PR to fix https://github.com/zeromq/pyzmq/issues/1167#issuecomment-637319447.

An alternative way would be using WeakSet for `self._sockets` but, because I'm not sure what side-effect becomes observable then, I follow original idea to use built-in set for `self._sockets`.


minrk:
I think WeakSet is indeed the right answer. Want to make that PR?

sakurai-youhei:
OK and Yes. Let me take some time to make the PR.

Issue #1378, opened 2020-05-08
Windows wheels missing for Python 3.5-3.6 (pyzmq 19.0.1)


Tobotimus:
Hi,

I just tried to install Jupyter Notebook on my Windows machine with Python 3.6 and got an unexpected error whilst trying to compile this library.

It looks as though the reason is, there's no wheel available for version 19.0.1 for Windows on Python 3.6 (or 3.5 for that matter). I'm not sure how they're deployed for this project and whether or not this is intentional, or a known issue. So I thought I'd ask :)

Regards,
Toby

omry:
Same issue, I had to pin my CI config to 19.0.0 to work around it.

minrk:
Takes a while for CI to upload the Windows wheels. They are up now. Thanks for checking!

omry:
@minrk, can this be improved?
zeromq is basic infrastructure for many projects and this kind of delay can potentially fail many projects continuous integration.
This is not the first time that I have to spend time figuring out why my CI run failed only to discover that was because of pyzeromq installation issues.

minrk:
Windows wheels are built on appveyor and uploaded manually. Uploading them automatically from CI would help a bunch, if you'd like to make a PR to appveyor.yml to add upload to PyPI.

A lot of the problem comes from building so many Windows wheels. Likely Python 3.5, at least, will stop being built at the next significant release. I hope to also stop building 32b Windows wheels for newer Pythons, which would cut the build number in half.

omry:
Unfortunately I don't have the cycles to create this PR, but I hope it's something you would consider.
for now I pinned my own CI to a specific version so this is not going to be an issue for me for a while.

minrk:
I don't either, so it likely won't be done until someone with time volunteers. I opened #1380 to track the task.

PR #1391, opened 2020-06-03
Simplified sockets tracking for reliable process teardown

sakurai-youhei:
This is PR to revise #1378.

Honestly speaking, because I found the test added through c6f414b fails with pyzmq==17.0.0 but succeeds with pyzmq==17.1.0, this PR presumably does nothing essential for #1167 but just strengthens foundation around at the #1167.

### DIFF
--- a/zmq/sugar/context.py
+++ b/zmq/sugar/context.py
@@ -7,11 +7,7 @@
 import atexit
 import os
 from threading import Lock
-import weakref
-
-# direct reference limits garbage collection issues
-# during process teardown
-weak_ref = weakref.ref
+from weakref import WeakSet
 
 from zmq.backend import Context as ContextBase
 from . import constants
@@ -46,7 +42,7 @@ class Context(ContextBase, AttributeSetter):
         else:
             self._shadow = False
         self.sockopts = {}
-        self._sockets = set()
+        self._sockets = WeakSet()
 
     def __del__(self):
         """deleting a Context should terminate it, without trying non-threadsafe destroy"""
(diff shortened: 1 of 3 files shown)


sakurai-youhei:
@minrk  Although I made a mess a bit, this PR is ready for review. May I have your feedback? Thanks.

zmq/sugar/context.py
```diff
@@ -46,10 +43,16 @@ def __init__(self, io_threads=1, **kwargs):
...
-        self._sockets = set()
+        self._sockets = WeakSet()
 
     def __del__(self):
         """deleting a Context should terminate it, without trying non-threadsafe destroy"""
+
+        # Referring sys module here conceals issue #1167 on at least CPython
+        # 3.5.4 on Windows. Without this, occurrence probability of the issue
+        # increases to approx. 30 ~ 50 % from 0 % (0 out of 240).
+        stdout.write("")
```
minrk, line 54: In some contexts, `sys.stdout` can be None or not writable, so this should either be checked for, exception caught, or another approach taken. Does the write itself have to execute, or is it just the reference to stdout, e.g. `getattr(stdout, 'write')`?

sakurai-youhei: Thank you. Exceptional cases (None or not writable) were totally out of my consideration. OK, let me try & seek a better option like you suggested.

sakurai-youhei: It's super wired but I found that `locals()` is sufficient. Because I suppose the `locals()` would be always safe, that's the best option I have ever imagined. If you have some idea, please let me know and I will try to make it one step ahead once again.

minrk: Weeeird...But if it works, should be safe. Thanks for digging in!

PR #1395, opened 2020-06-15
Improve Poll documentation

rogerthat94:
I found the explanation of the return value for zmq.Socket.poll() to be confusing. I changed the documentation to try to eliminate this confusion. In particular, the documentation for the return value from zmq.Poller.poll() uses the word `events`, so it seemed better to refer to the value from zmq.Poller.poll() as `event`.

I also added some clarification to the documentation for zmq.Poller.poll() explaining how the timeout works. I realize this is obvious to someone familiar with the POSIX standards for select and poll. However, pyzmq might be used by people who have not done low level programming on Linux or Unix, and I feel this additional information will be useful for them.

### DIFF
--- a/zmq/sugar/poll.py
+++ b/zmq/sugar/poll.py
@@ -78,7 +78,7 @@ class Poller(object):
         """Poll the registered 0MQ or native fds for I/O.
         
         If there are currently events ready to be processed, this function will return immediately.
-        Otherwise, this function will return as soon as events are available or after timeout 
+        Otherwise, this function will return as soon the first event is available or after timeout 
         milliseconds have elapsed.
 
         Parameters
--- a/zmq/sugar/poll.py
+++ b/zmq/sugar/poll.py
@@ -91,10 +91,10 @@ class Poller(object):
         -------
         events : list of tuples
             The list of events that are ready to be processed.
-            This is a list of tuples of the form ``(socket, event)``, where the 0MQ Socket
+            This is a list of tuples of the form ``(socket, event_mask)``, where the 0MQ Socket
             or integer fd is the first element, and the poll event mask (POLLIN, POLLOUT) is the second.
             It is common to call ``events = dict(poller.poll())``,
-            which turns the list of tuples into a mapping of ``socket : event``.
+            which turns the list of tuples into a mapping of ``socket : event_mask``.
         """
         if timeout is None or timeout < 0:
             timeout = -1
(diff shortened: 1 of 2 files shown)


zmq/sugar/socket.py
```diff
@@ -677,7 +677,7 @@ def recv_json(self, flags=0, **kwargs):
     _poller_class = Poller
 
     def poll(self, timeout=None, flags=POLLIN):
-        """Poll the socket for events.
+        """Poll the socket for an event.
```
minrk, line 680: Can we leave this plural? It is a mask reporting potentially multiple events, so it seems confusing to me to say "event" refers to one or more events that are ready.

zmq/sugar/poll.py
```diff
@@ -76,6 +76,10 @@ def unregister(self, socket):
 
     def poll(self, timeout=None):
         """Poll the registered 0MQ or native fds for I/O.
+        
+        If there are currently events ready to be processed, this function will return immediately.
+        Otherwise, this function will return as soon as events are available or after timeout 
+        milliseconds have elapsed.
```
minrk, line 82: 👍

zmq/sugar/socket.py
```diff
@@ -690,9 +690,9 @@ def poll(self, timeout=None, flags=POLLIN):
 
         Returns
         -------
-        events : int
-            The events that are ready and waiting,
-            0 if the timeout was reached with no events.
+        event : int
```
minrk, line 693:
```suggestion
        events : int (bitmask)
            The poll event(s) (POLLIN, POLLOUT) that are ready,
            if any. Use `events & POLLIN` or `events & POLLOUT` to check
            which events are ready.
            0 if timeout is reached before any events.
```

zmq/sugar/socket.py
rogerthat94: The [documentation](https://github.com/zeromq/pyzmq/blob/d28d7cae78a1292bf882924d757cdf884a61eef6/zmq/sugar/poll.py#L94) for Poller.poll() refers to this bitmask as `event`. It seems like these should be consistent, to reduce confusion. 

Would it be better to change the Poller.poll() to call this bitmask `events`?

minrk: You're definitely right they should match! I do think plural makes more sense.

rogerthat94: I changed both to event_mask. Does that sound reasonable? When I first read the document, I thought events was a list of events. When I realized it was an int, I thought it referred to the number of events. 

I think we should make it really clear that this is actually the bitmap.

minrk: Good idea, thanks!

PR #1404, opened 2020-07-17
WIP: Added arm64 jobs in Travis-ci

### LINKED ISSUE #1401: Wheel support for aarch64.
**Summary**
Installing pyzmq on aarch64 via pip using command "pip3 install pyzmq" tries to build wheel from source code

**Problem description**
pyzmq don't have wheel for aarch64 on PyPI repository. So, while installing pyzmq via pip on aarch64, pip builds wheel for same resulting in it takes more time to install pyzmq. Making wheel available for aarch64 will benefit aarch64 users by minimizing pyzmq installation time.

**Expected Output**
Pip should be able to download pyzmq wheel from PyPI repository rather than building it from source code.

@pyzmq-team, please let me know if I can help you building wheel/uploading to PyPI repository. I am curious to make pyzmq wheel available for aarch64. It will be a great opportunity for me to work with you.

odidev:
Updated travis.yml file to add ARM64 jobs in Travis-CI. Added support for python version 3.6, 3,7 and 3.8 for ARM64 jobs.
Resolves #1401.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,7 @@ before_install:
       export ZMQ_DRAFT_API=1
     fi
   - pip install -r test-requirements.txt
+  - pip install invoke requests
   - |
     # install pinned tornado
     if [[ "$TORNADO" == "master" ]]; then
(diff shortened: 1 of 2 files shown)


minrk:
This is great! I think we can do fewer arm versions, just ZMQ= and ZMQ=bundled for python 3.7, 3.8 (4 total) and skip the tornado matrix items, which have no reason to behave differently on arm.

Since this is new, I think we can limit oldest ARM builds to 3.7, unless there's a really strong case that old 3.6 is very widely used on ARM.

For the wheel builds, I think we want to port over the logic from tools/tasks.py, which builds libzmq+libsodium from source, and we want to make sure the linux wheels built on CI do the same. Also for the wheel builds, we probably want to run the tests after installing from the wheel instead of running tests from the source directory prior to building an untested wheel.

odidev:
> I think we can do fewer arm versions, just ZMQ= and ZMQ=bundled for python 3.7, 3.8 (4 total), and skip the tornado matrix items, which have no reason to behave differently on arm.

@minrk As per your suggestion I have updated .travis.yml to run for python 3.7, 3.8 (4 total), and removed all the other jobs for arm64.

> I think we want to port over the logic from tools/tasks.py, which builds libzmq+libsodium from source, and we want to make sure the Linux wheels built on CI do the same.

Also, I have updated tools/tasks.py to create wheel for aarch64, now wheels are getting created successfully.

Please have a look and share your opinion.

PR #1417, opened 2020-09-08
Require Python >=3.5

minrk:
Pythons 2.7-3.5 are EOL, but 3.5 is still the default python3 in a supported Ubuntu LTS for one more year

PR #1419, opened 2020-09-09
bump bundled libzmq to 4.3.3

### LINKED ISSUE #1418: Problem: TestAsyncioAuthentication::test_blacklist hangs with latest libzmq release
In libzmq 4.3.3 CVE-2020-15166 ( https://github.com/zeromq/libzmq/security/advisories/GHSA-25wp-cf8g-938m ) was fixed. Previously, unauthenticated peers would still connect, and could affect servers and authenticated clients, even though messages would not be exchanged on the wire. Malicious clients could use this for denial of service attacks.

It looks like the test_blacklist is relying on this behaviour, as it expects a ZMQ_PUSH send to succeed, even though there is no authenticated peer connected, so with the fix it now hangs forever in https://github.com/zeromq/pyzmq/blob/master/zmq/tests/asyncio/_test_asyncio.py#L466

Yes, this is a change in existing behaviour - but it was really broken, so there's no alternative unfortunately.

minrk:
fix asyncio can_connect test to wait for POLLOUT

threaded auth tests already includes this check

closes #1418

### DIFF
--- a/buildutils/bundle.py
+++ b/buildutils/bundle.py
@@ -32,14 +32,14 @@ pjoin = os.path.join
 # Constants
 #-----------------------------------------------------------------------------
 
-bundled_version = (4, 3, 2)
+bundled_version = (4, 3, 3)
 vs = '%i.%i.%i' % bundled_version
 libzmq = "zeromq-%s.tar.gz" % vs
 libzmq_url = "https://github.com/zeromq/libzmq/releases/download/v{vs}/{libzmq}".format(
     vs=vs,
     libzmq=libzmq,
 )
-libzmq_checksum = "sha256:ebd7b5c830d6428956b67a0454a7f8cbed1de74b3b01e5c33c5378e22740f763"
+libzmq_checksum = "sha256:9d9285db37ae942ed0780c016da87060497877af45094ff9e1a1ca736e3875a2"
 
 HERE = os.path.dirname(__file__)
 ROOT = os.path.dirname(HERE)
--- a/buildutils/bundle.py
+++ b/buildutils/bundle.py
@@ -39,7 +39,9 @@ libzmq_url = "https://github.com/zeromq/libzmq/releases/download/v{vs}/{libzmq}"
     vs=vs,
     libzmq=libzmq,
 )
-libzmq_checksum = "sha256:9d9285db37ae942ed0780c016da87060497877af45094ff9e1a1ca736e3875a2"
+libzmq_checksum = (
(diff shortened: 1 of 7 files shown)


bluca (approved):

minrk:
Building bundled 4.3.3 apparently needs some new compilation options (disable gnutls, ws transport, etc.), which will take a bit of investigating. Probably should have called it at least 4.4, since there are lots of new features, new dependencies, etc.

bluca:
> Building bundled 4.3.3 apparently needs some new compilation options (disable gnutls, ws transport, etc.), which will take a bit of investigating. Probably should have called it at least 4.4, since there are lots of new features, new dependencies, etc.

These are all disabled by default though? Unless you are enabling drafts of course

bluca:
> > Building bundled 4.3.3 apparently needs some new compilation options (disable gnutls, ws transport, etc.), which will take a bit of investigating. Probably should have called it at least 4.4, since there are lots of new features, new dependencies, etc.
> 
> These are all disabled by default though? Unless you are enabling drafts of course

If you want to enable drafts but disable WS, there's even a separate flag: --enable-ws (default is the value of --enable-draft)

minrk:
Yeah, the tricky part is that the bundled libzmq can't be built by `configure/make/make install`, so when there are new required defines, etc. they need to be added manually, either by regenerating platform.hpp with `./configure` or adding the defines in setup.py, depending on how they are managed. I'll look into this when I get a chance if someone else doesn't get to it first, probably some time next month.

I hope there aren't any changes that will break the ancient VC 9 build on Windows, required for Python 2. If there are, I'll probably just cut it here and say we've shipped our last version supporting Python 2, which I was planning to do soon, anyway.

bluca:
> Yeah, the tricky part is that the bundled libzmq can't be built by `configure/make/make install`, so when there are new required defines, etc. they need to be added manually, either by regenerating platform.hpp with `./configure` or adding the defines in setup.py, depending on how they are managed. I'll look into this when I get a chance if someone else doesn't get to it first, probably some time next month.
> 
> I hope there aren't any changes that will break the ancient VC 9 build on Windows, required for Python 2. If there are, I'll probably just cut it here and say we've shipped our last version supporting Python 2, which I was planning to do soon, anyway.

The no-draft build works on 20 years old Solaris (I have a VM), so it should be safe

minrk:
looks like there are changes that prevent building with py27 on Windows with VC9, given the syntax errors. Will wait for #1417 to drop support for python 2, then can come back to this.

PR #1421, opened 2020-09-13
Check PKG_CONFIG env variable when building

kapsh:
Build systems traditionally refer to PKG_CONFIG environment variable with fallback to `pkg-config` executable if no such set.


minrk:
Makes sense, thanks

PR #1422, opened 2020-09-13
Use TestCase.assertTrue instead of deprecated assert_ alias

kapsh:
Less warnings is better. 

https://docs.python.org/3/library/unittest.html#deprecated-aliases


zmq/tests/test_message.py
```diff
@@ -114,8 +114,8 @@ def test_lifecycle1(self):
             self.assertEqual(s, b(str(m)))
             self.assertEqual(s, bytes(m2))
             self.assertEqual(s, m.bytes)
-            # self.assert_(s is str(m))
-            # self.assert_(s is str(m2))
+            # self.assertTrue(s is str(m))
```
kapsh, line 117: Weird, but I'll leave it as is.

minrk:
Thanks! Since we use pytest now, these could also all be bare `assert expr` instead of methods.

kapsh:
Sure, but it's more consistent as long as everything else refers to unittest methods.

PR #1426, opened 2020-10-02
Update examples/security/asyncio-ironhouse.py

### LINKED ISSUE #1254: Update examples/security/asyncio-ironhouse.py
Hello,

I'm looking at [asyncio-ironhouse.py](https://github.com/zeromq/pyzmq/blob/3a2fe2f8fef1f004104d94b744bd183b74053af8/examples/security/asyncio-ironhouse.py)

- it uses ZMQEventLoop, but that is deprecated
- other minor issues:
  - it uses older syntax - `@asyncio.coroutine` instead of `async def`, and `yield from` instead of `await`
  - I would use `pathlib.Path` instead of `os.path.*`
  - `async with` could be used to ensure things are closed properly
  - `asyncio.run()` could be used instead of `loop.run_until_complete()` (if it's ok that it's only Python 3.7+)

I would send a pull request with these changes. What do you think?

Petr M.

messa:
Closes #1254 – I'm sorry it took so long 😇 

The one thing I wasn't sure about was whether use context managers for closing the socket and stopping the auth thread. It would make the example more complicated.


minrk:
Thanks!

Issue #1436, opened 2020-10-19
Idea - Add "bound" and "connected" context managers to Socket


randallpittman:
I find it important at times to connect/disconnect a socket to an subscriber endpoint in an ephemeral manner. I made a simple subclass of Socket to do this in a `with` block:  https://gitlab.com/osu-nrsg/zmq-quicklink/-/blob/master/zmq_quicklink/__init__.py

Happy to do a PR if folks think this would be worth including in pyzmq.

minrk:
Yeah, makes sense! I think you can make these context managers returned from the existing `bind`/`connect` methods without a real downside, I think. If you can think of a clear downside, then new methods are okay, too.

PR #1439, opened 2020-10-20
Make Socket connect() and bind() context managers

randallpittman:
My attempt at addressing #1436 as suggested not adding new methods but extending `connect()` and `bind()` to make them return context managers that automatically disconnect/unbind.

~~I couldn't use `contextlib.contextmanager` because that wouldn't do the connecting or binding when used outside of a `with` block, so instead I created a small connect/bind context manager class and turned `connect` and `bind` into factory functions. I copied (and amended) the docstrings for `connect` and `bind` from their respective Cython versions.~~

### DIFF
--- a/zmq/sugar/socket.py
+++ b/zmq/sugar/socket.py
@@ -131,9 +131,9 @@ class Socket(SocketBase, AttributeSetter):
             self.addr = addr
             self.bind = bind
             if self.bind:
-                SocketBase.bind(self.sock, addr)
+                SocketBase.bind(self.sock, self.addr)
             else:
-                SocketBase.connect(self.sock, addr)
+                SocketBase.connect(self.sock, self.addr)
 
         def __enter__(self):
             return self.sock
--- a/zmq/sugar/socket.py
+++ b/zmq/sugar/socket.py
@@ -9,6 +9,7 @@ import errno
 import random
 import sys
 import warnings
+from contextlib import contextmanager
 
 import zmq
 from zmq.backend import Socket as SocketBase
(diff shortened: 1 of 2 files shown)


randallpittman:
I meant to do a separate PR with the updated implementation, but in any case I think it shows clearer intent of what's going on.

zmq/sugar/socket.py
```diff
@@ -120,6 +121,66 @@ def close(self, linger=None):
...
+
+        Parameters
+        ----------
+        addr : str
+            The address string. This has the form 'protocol://interface:port',
+            for example 'tcp://127.0.0.1:5555'. Protocols supported include
+            tcp, udp, pgm, epgm, inproc and ipc. If the address is unicode, it is
+            encoded to utf-8 first.
+        """
+        SocketBase.bind(self, addr)
```
minrk, line 163: We can use super to call the parent method:

```suggestion
        super().bind(addr)
```

zmq/sugar/socket.py
```diff
@@ -120,6 +121,66 @@ def close(self, linger=None):
...
+        return self._bind_cm(addr)
+
+    def connect(self, addr):
+        """s.connect(addr)
+
+        Connect to a remote 0MQ socket.
+
+        Returns a context manager that may be used to automatically disconnect in a `with` block.
+
```
minrk, line 172: Can you add `.. versionadded:: 20.0` to the docstrings?

minrk (commented):
Great! A couple of comments in-line, but looks fine to me. Thank you!

zmq/sugar/socket.py
randallpittman: Sure. I'm sometimes hesitant to use `super()` when there's multiple inheritance (possibly due to problems with my understanding of multiple inheritance), but it's probably safe here.

zmq/sugar/socket.py
randallpittman: Yup.

randallpittman:
@minrk Updates pushed.

randallpittman:
@minrk I didn't think until now to add these to the asyncio versions (just now learning async zmq) but would not mind doing so. Or I can do that in a separate PR. Let me know.

minrk:
Thanks! Since bind/connect are not themselves async methods, I don't think there's a benefit to supporting `async with`, so this should be all we need, I think.

PR #1446, opened 2020-11-24
run all tests on github actions

minrk:
travis is effectively shutting down its open source program, appveyor is super slow

this significantly reduces the Windows test matrix since wheels are built in a separate workflow.


minrk:
I think PRs don't run new workflows by default, so these can be watched over [here](https://github.com/minrk/pyzmq/actions?query=branch%3Aactions)

kinow:
> I think PRs don't run new workflows by default, so these can be watched over [here](https://github.com/minrk/pyzmq/actions?query=branch%3Aactions)

I think it does only if the branch is from the same repository (`zeromq`) and not from a fork.

PR #1466, opened 2020-12-30
Introducing load meta data option for load_certificate method

nilsmelchert:
Figured, that the option to extract the metadata from a certificate was missing. Merging this should not cause any compatibility issues and extends the usablity of certificate metadata.

NOTE: This is only compatible with Python >=3.8, since the [Walrus Operator](https://stackoverflow.com/questions/50297704/syntax-and-assignment-expressions-what-and-why) was used. If you wish more compatibility, I can push those changes to this PR too.

### DIFF
--- a/zmq/auth/certs.py
+++ b/zmq/auth/certs.py
@@ -107,10 +107,11 @@ def load_certificate(filename, load_metadata=False):
             if line.startswith(b'#'):
                 continue
             if load_metadata and line.startswith(b'metadata'):
-                while not (line := next(f)).startswith(b'curve'):
-                    k = line.split(b"=", 1)[0].strip(b' \t\'"\r\n').decode()
-                    v = line.split(b"=", 1)[1].strip(b' \t\'"\r\n').decode()
+                line = next(f)
+                while not line.startswith(b'curve'):
+                    k, v = [element.strip(b' \t\'"\r\n').decode() for element in line.split(b"=", 1)]
                     metadata[k] = v
+                    line = next(f)
             if line.startswith(b'public-key'):
                 public_key = line.split(b"=", 1)[1].strip(b' \t\'"')
             if line.startswith(b'secret-key'):
--- a/zmq/auth/certs.py
+++ b/zmq/auth/certs.py
@@ -82,18 +82,22 @@ def create_certificates(key_dir, name, metadata=None):
     return public_key_file, secret_key_file
 
 
-def load_certificate(filename):
-    """Load public and secret key from a zmq certificate.
+def load_certificate(filename, load_metadata=False):
+    """Load public and secret key from a zmq certificate. Optionally, also the metadata
+    from the key is returned.
 
-    Returns (public_key, secret_key)
+    Returns (public_key, secret_key, [metadata])
 
     If the certificate file only contains the public key,
     secret_key will be None.
+    If there is not metadata in the certificate,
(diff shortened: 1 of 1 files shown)


nilsmelchert:
Since some checks did non pass in 74d2bba. Therefore, the Walrus Operator was replaced in 70a8177 to ensure backward compatibility for earlier python versions.

zmq/auth/certs.py
```diff
@@ -102,6 +106,12 @@ def load_certificate(filename):
             line = line.strip()
             if line.startswith(b'#'):
                 continue
+            if load_metadata and line.startswith(b'metadata'):
```
minrk, line 109: Is it valid for a file to define metadata after the keys? If so, this won't always work. Can you also add some tests?

nilsmelchert: Thats indeed true. If the meta data is defined after the key, this won't work.

However, as far as I know, every certificate, which is created by zeromq has the same format:

```
metadata
    my_metadata1 = "172.17.64.145"
    my_metadata2 = "192.168.56.1"
    my_metadata3 = "130.75.27.242"
curve
    public-key = "2H0gQQr{^i6BUL:re4H{pb)b+LI:qtMMYvvl}1yT"
    secret-key = "YShO$Irv2}WM&jow1Kz.Oq$2!B6P4?f?/v^LYe<"
```

Am I wrong about this? If so, I will fix it. Do you want me to add tests anyways?

minrk: That may be true in practice, but [the spec](https://rfc.zeromq.org/spec/4/) doesn't indicate an order is required. It might be appropriate to add a generic ZPL parser if we are starting to consume more than one section field, then call that `read_zpl(path)` in `load_cert`. What do you think?

nilsmelchert: I like the idea of a generic ZPL parser. I will design one and introduce it either here, or in another PR.

What do you think, where we should place such a parser within this repo?

minrk: Let's put it in utils as a private utility function, e.g. `utils.zpl.read_zpl(path) -> dict` then call it here.

nilsmelchert: Sounds good. I will close this PR for now and create a new one, when I am done.

nilsmelchert: There is actually already the python package [pyzpl](https://github.com/mbarkhau/pyzpl) on PyPi with under MIT license with that kind of functionality. We might either just use some of that code or merge that repository to pyzmq. What do you think?

minrk: Hm, knowing that pyzpl exists and assuming it works, maybe the best thing to do is point people to that if they want more detailed loading of more fields instead of integrating it here?

nilsmelchert: The pyzpl package seems to be well maintained and does what it is supposed to do. I tested it for my own needs.

However, I think this kind of functionality should be available in the "official" pyzmq package. But that is just me.

minrk: That's a good point. For now, since pyzmq has no internal support for using information in fields other than the keys, let's point folks to pyzpl for reading zpl full files. We can revisit if there's a case to be made for support using metadata or other fields in the existing auth methods, etc.

nilsmelchert: Alright. Then I will close this issue for now.

PR #1479, opened 2021-01-22
specify default value of send/recv flags in type stubs

vinnik-dmitry07:
https://pyzmq.readthedocs.io/en/latest/api/zmq.html#zmq.Socket.send
https://pyzmq.readthedocs.io/en/latest/api/zmq.html#zmq.Socket.recv


minrk:
Thanks!

Issue #1477, opened 2021-01-20
pep561 compliance


arnimarj:
Hi. Just noticed that the package has started shipping type stubs (`pyi` extension). When trying to test this with my current projects I was unable to make mypy find the stubs. It seems that `py.typed` is missing. Its existence is a signal that the package is partially stubbed:

https://www.python.org/dev/peps/pep-0561/#partial-stub-packages

See https://github.com/maxmind/GeoIP2-python/pull/93/files for a similar fix in `geoip2`

If/when the project ships inline annotations this becomes unnecessary, I think.

minrk:
Makes sense, thanks! Want to make the appropriate PR?

How did you discover that it didn't work? It would be handy to have a test, but everything I tried worked fine.

arnimarj:
> Makes sense, thanks! Want to make the appropriate PR?
> 
> How did you discover that it didn't work? It would be handy to have a test, but everything I tried worked fine.

Hi. We have been maintaining our own set of stubs to help with the development of some internal libraries, and the test was simply to remove those stubs. I'll try to steal some time for this in the next few days. Cheers.

arnimarj:
> Makes sense, thanks! Want to make the appropriate PR?
> 
> How did you discover that it didn't work? It would be handy to have a test, but everything I tried worked fine.

Hi. I just opened a PR here: https://github.com/zeromq/pyzmq/pull/1483

PR #1483, opened 2021-01-23
Feature/pep 561

arnimarj:
In order to make the new stubs discoverable by mypy, the package needs to include the `py.typed` file marker.

I also removed default values from the stubs. They sometimes make sense, e.g. when using `typing.overload` (see e.g. https://github.com/python/typeshed/blob/master/third_party/2and3/redis/client.pyi#L96) but not here I think.

Fixes https://github.com/zeromq/pyzmq/issues/1477


minrk:
Thanks!

PR #1485, opened 2021-01-25
[windows] keep pypy windows wheels with libzmq as an Extension

minrk:
pypy depends on external  vcruntime-redist install, unlike CPython

delvewheel bundling doesn't seem to work. Still unclear why


minrk:
Hm, extremely weird that this worked once on CI, but isn't working since then. I cannot imagine what would be the cause, then.

PR #1492, opened 2021-01-28
Add python 3.6 wheel support for pyzmq

odidev:
Added python 3.6 wheel support for pyzmq.


minrk:
What use case prompted this? Since these wheels are all new and the aarch builds are very expensive, I thought maybe we didn't need to go back to soon-to-be out-of-date Python 3.6.

rhenwood-arm:
Hi @minrk (I work with @odidev) I've been advised that this package is valuable for folks using Graviton2. I have a colleague who works with Travis-CI - let me see if I can get some help from them.

PR #1497, opened 2021-01-29
don't omit vcruntime140.dll from windows wheels

minrk:
maybe this is why it doesn't work on conda?


minrk:
Nope, tha'ts not it. #1498

PR #1524, opened 2021-05-09
try piggy-backing on tornado for proactor loop support

### LINKED ISSUE #1423: Asyncio error on Windows Python 3.8 due to proactor event loop
Using `zmq.asyncio` on Windows with Python 3.8 fails, due to `asyncio` using the "proactor" event loop by default. See https://github.com/tornadoweb/tornado/issues/2608 for related discussion.

Specifying that `asyncio` should use the selector event loop (the previous default) allows `zmq.asyncio` to run successfully.

Can pyzmq be updated to support the proactor event loop, or do we need to manually change the event loop to selector for anything using `zmq.asyncio`?

Example: the code below generates an error unless the `asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())` line is uncommented.

```python
import asyncio
import zmq
import zmq.asyncio

port = 5556

#asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
ctx = zmq.asyncio.Context()


async def recv_and_process():
    sock = ctx.socket(zmq.REP)
    sock.connect('tcp://localhost:%d' % port)
    msg = await sock.recv_string() 
    await asyncio.sleep(1)
    await sock.send_string('ack')


async def send_and_wait():
    sock = ctx.socket(zmq.REQ)
    sock.bind('tcp://*:%d' % port)
    await sock.send_string('msg')
    reply = await sock.recv_string()


async def sendAndRecv():
    await asyncio.gather(
        send_and_wait(),
        recv_and_process()
    )


asyncio.run(sendAndRecv())
```

libzmq-4.3.2
pyzmq-19.0.2
Python-3.8.5 (tags/v3.8.5:580fbb0, Jul 20 2020, 15:57:54) [MSC v.1924 64 bit (AMD64)]

### LINKED ISSUE #1521: Using zmq.asyncio, getting NotImplementedError when creating a REQ socket
Had a simple REQ-REP application working fine. I went to convert the python client code to asyncio and ran into a NotImplemented error. 

I tried to find it. Does this mean that AbstractEventLoop virtual methods are getting called?

pyzmq 22.0.3
python 3.9
win10

### Traceback
```
  File "..\minimal_example.py", line 33, in <module>
    asyncio.run(main())
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.9_3.9.1264.0_x64__qbz5n2kfra8p0\lib\asyncio\runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.9_3.9.1264.0_x64__qbz5n2kfra8p0\lib\asyncio\base_events.py", line 642, in run_until_complete
    return future.result()
  File "..\minimal_example.py", line 27, in main
    bl_1300 = ServerConnection(addr='127.0.0.1', port=9999)
  File "..\minimal_example.py", line 21, in __init__
    self._zmq_socket = _zmq_context.socket(zmq.REQ)
  File "..\venv\lib\site-packages\zmq\sugar\context.py", line 235, in socket
    s = self._socket_class(self, socket_type, **kwargs)
  File "..\venv\lib\site-packages\zmq\_future.py", line 150, in __init__
    self._init_io_state()
  File "..\venv\lib\site-packages\zmq\asyncio.py", line 51, in _init_io_state
    self.io_loop.add_reader(self._fd, lambda: self._handle_events(0, 0))
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.9_3.9.1264.0_x64__qbz5n2kfra8p0\lib\asyncio\events.py", line 504, in add_reader
    raise NotImplementedError
NotImplementedError
Exception ignored in: <function Socket.__del__ at 0x000002ECFB11C280>
  File "..\venv\lib\site-packages\zmq\sugar\socket.py", line 66, in __del__
  File "..\venv\lib\site-packages\zmq\_future.py", line 168, in close
  File "..\venv\lib\site-packages\zmq\asyncio.py", line 58, in _clear_io_state
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.9_3.9.1264.0_x64__qbz5n2kfra8p0\lib\asyncio\events.py", line 507, in remove_reader
NotImplementedError: 

Process finished with exit code 1
```

### minimal_example.py
```python
#minmal_example.py
#Using zmq.asyncio, NotImplementedError when creating a REQ socket

import asyncio
import zmq
import zmq.asyncio

_zmq_context = None


class ServerConnection:
    """Represents a remote system, running a zmq server."""
    _zmq_socket = None

    def __init__(self, addr='127.0.0.1', port=9999):
        """Contructor. Supply the zmq address string, addr, to reach this endstation."""

        global _zmq_context

        # the first server object will create the global zmq context
        if not _zmq_context:
            _zmq_context = zmq.asyncio.Context()

        self._zmq_socket = _zmq_context.socket(zmq.REQ)

        # ^^^^^^^^^^^^^^ fails NotImplementedError


async def main():
    bl_1300 = ServerConnection(addr='127.0.0.1', port=9999)

    print(await bl_1300.request('server status', {}))


if __name__ == '__main__':
    asyncio.run(main())
```

minrk:
Tornado 6.1 enables support for proactor by running a separate selector loop in a thread

<del>try piggy-backing on that functionality by using tornado's AddThreadEventLoop when someone attempts to use zmq.asyncio with proactor</del>

Went with vendoring SelectorThread from https://github.com/tornadoweb/tornado/pull/3029 so no dependency is added.

closes #1521
closes #1423

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -149,7 +149,8 @@ jobs:
         if: startsWith(matrix.os, 'ubuntu')
         run: |
           sudo apt-get update
-          sudo apt-get install -y -qq libzmq3-dev libsodium-dev
+          sudo apt-get -y remove libzmq5 || true # workaround https://github.com/actions/virtual-environments/issues/3317
+          sudo apt-get -y install libzmq3-dev libsodium-dev
 
       - name: install libzmq-dev
         if: matrix.zmq == 'head'
--- a/zmq/_asyncio_selector.py
+++ b/zmq/_asyncio_selector.py
@@ -90,6 +90,7 @@ class SelectorThread:
             None
         )  # type: Optional[Tuple[List[_FileDescriptorLike], List[_FileDescriptorLike]]]
         self._closing_selector = False
+        self._closed = False
         self._thread = threading.Thread(
             name="Tornado selector",
             daemon=True,
@@ -121,6 +122,8 @@ class SelectorThread:
         self._waker_w.close()
 
     def close(self) -> None:
+        if self._closed:
+            return
         with self._select_cond:
             self._closing_selector = True
             self._select_cond.notify()
@@ -129,6 +132,7 @@ class SelectorThread:
         _selector_loops.discard(self)
         self._waker_r.close()
         self._waker_w.close()
+        self._closed = True
 
     def _wake_selector(self) -> None:
         try:
--- a/zmq/_asyncio_selector.py
+++ b/zmq/_asyncio_selector.py
@@ -1,0 +1,242 @@
+"""Ensure asyncio selector methods (add_reader, etc.) are available
+
+tornado 6.1 adds AddThreadSelectorEventLoop event loop,
+running select in a thread and defining these methods on the running event loop.
+
+This factors out the functionality of AddThreadSelectorEventLoop
+into a standalone SelectorThread object which can be attached to any running event loop.
+
+Vendored from tornado v6.1.0-64-g289c834b (PR #3029)
+
+Redistributed under license Apache-2.0
+"""
+
+import asyncio
+import atexit
+import errno
+import functools
+import select
+import socket
+import threading
+import typing
+from typing import (
+    Any,
+    TypeVar,
+    Awaitable,
+    Callable,
+    Union,
+    Optional,
+    List,
+    Tuple,
+    Dict,
+)  # noqa: F401
+
+if typing.TYPE_CHECKING:
+    from typing import Set  # noqa: F401
+    from typing_extensions import Protocol
+
+    class _HasFileno(Protocol):
+        def fileno(self) -> int:
+            pass
+
+    _FileDescriptorLike = Union[int, _HasFileno]
+
+
+# Collection of selector thread event loops to shut down on exit.
+_selector_loops = set()  # type: Set[SelectorThread]
+
+
+def _atexit_callback() -> None:
+    for loop in _selector_loops:
+        with loop._select_cond:
+            loop._closing_selector = True
+            loop._select_cond.notify()
+        try:
+            loop._waker_w.send(b"a")
+        except BlockingIOError:
+            pass
+        # If we don't join our (daemon) thread here, we may get a deadlock
+        # during interpreter shutdown. I don't really understand why. This
+        # deadlock happens every time in CI (both travis and appveyor) but
+        # I've never been able to reproduce locally.
+        loop._thread.join()
+    _selector_loops.clear()
+
+
+atexit.register(_atexit_callback)
+
+
+class SelectorThread:
+    """Define ``add_reader`` methods to be called in a background select thread.
+
+    Instances of this class start a second thread to run a selector.
+    This thread is completely hidden from the user;
+    all callbacks are run on the wrapped event loop's thread.
+
+    Typically used via ``AddThreadSelectorEventLoop``,
+    but can be attached to a running asyncio loop.
+    """
+
+    def __init__(self, real_loop: asyncio.AbstractEventLoop) -> None:
+        self._real_loop = real_loop
+
+        # Create a thread to run the select system call. We manage this thread
+        # manually so we can trigger a clean shutdown from an atexit hook. Note
+        # that due to the order of operations at shutdown, only daemon threads
+        # can be shut down in this way (non-daemon threads would require the
+        # introduction of a new hook: https://bugs.python.org/issue41962)
+        self._select_cond = threading.Condition()
+        self._select_args = (
+            None
+        )  # type: Optional[Tuple[List[_FileDescriptorLike], List[_FileDescriptorLike]]]
+        self._closing_selector = False
+        self._thread = threading.Thread(
+            name="Tornado selector",
+            daemon=True,
+            target=self._run_select,
+        )
+        self._thread.start()
+        # Start the select loop once the loop is started.
+        self._real_loop.call_soon(self._start_select)
+
+        self._readers = {}  # type: Dict[_FileDescriptorLike, Callable]
+        self._writers = {}  # type: Dict[_FileDescriptorLike, Callable]
+
+        # Writing to _waker_w will wake up the selector thread, which
+        # watches for _waker_r to be readable.
+        self._waker_r, self._waker_w = socket.socketpair()
+        self._waker_r.setblocking(False)
+        self._waker_w.setblocking(False)
+        _selector_loops.add(self)
+        self.add_reader(self._waker_r, self._consume_waker)
+
+    def __del__(self) -> None:
+        # If the top-level application code uses asyncio interfaces to
+        # start and stop the event loop, no objects created in Tornado
+        # can get a clean shutdown notification. If we're just left to
+        # be GC'd, we must explicitly close our sockets to avoid
+        # logging warnings.
+        _selector_loops.discard(self)
+        self._waker_r.close()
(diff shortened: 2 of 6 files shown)


minrk:
You can try calling `asyncio.set_event_loop_policy(WindowsSelectorEventLoopPolicy())` before invoking any asyncio methods to see if that helps. That would indicate this change is relevant.

Jeducious:
Thanks! I did try that, results were that the warning does go away at least. Still having issues where the test will progress for a while but then hangs eventually. I can't 100% say for sure the its due to this though, I am also having problems on linux too, so for the moment, I can't really prove that the hang is due to the event loop policy. 

If that changes I'll report back.

minrk:
If changing the policy still hangs, then I think it's probably not that, and something else, possibly related to edge-triggering issues. These things can be hard to track down!

Jeducious:
Indeed! I am digging, the problem is that it's difficult to reproduce reliably. There are other things in here than pyzmq, for example the python logging module. I am currently removing all logging to check that this is not a factor. 

So I'm proceeding to eliminate things by removing them where I can. Will let you know if anything points back at ZMQ.

Jeducious:
@minrk 

Ok, have a question I am seeing an error on linux now. This error suggests I am exhausting the file descriptor quota. I had a look at the offending process, looks like it is accumulating fd's alright, but I wondered if you'd be able to tell me if this looks like something the asyncio pyzmq sockets might use? The majority of the fd's in use are of the type eventfd. 

Man page on that is [here](https://man7.org/linux/man-pages/man2/eventfd.2.html)

It basically says that these are used an event wait/notify mechanism by user-space applications, so I am guessing that this either

1. Asyncio tasks doing this
2. Pyzmq sockets... maybe?
3. Something else entirely that I am missing (catchall, had to throw that in there to cover my ignorance).

It seems like they are not being released, but can't confirm. The process hung, so they might have been if it had closed :)

```
python3 33202 ubuntu   60u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   61u  a_inode               0,14        0  10299 [eventpoll]
python3 33202 ubuntu   62u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   63u  a_inode               0,14        0  10299 [eventpoll]
python3 33202 ubuntu   64u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   65u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   66u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   67u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   68u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   69u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   70u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   71u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   72u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   73u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   74u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   75u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   76u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   77u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   78u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   79u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   80u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   81u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   82u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   83u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   84u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   85u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   86u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   87u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   88u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   89u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   90u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   91u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   92u  a_inode               0,14        0  10299 [eventfd]
python3 33202 ubuntu   93u  a_inode               0,14        0  10299 [eventfd]
```

minrk:
Certainly possible, but I can't be sure. I don't know exactly what operations create these.

You might check `asyncio.all_tasks()` to see all the asyncio tasks you have running.

It's conceivable you have launched some task/future that you lose track of without awaiting or cancelling. This could be due to your code, or even a pyzmq bug.

Jeducious:
Ok, I think I am getting close (though honestly concurrent programming can certainly prove me wrong it seems)

I have several workers which each run as an asyncio task. Each worker has a zmq.DEALER socket, plus, I create a monitor socket for each of the dealers using get_monitor_socket.

During a shutdown I call cancel on each worker task, this triggers a shutdown handler which calls the disable_monitor() method for the dealer socket. This is where the loop hangs. 

It seems a little bit random as sometimes a few workers will all be able to cleanly shutdown, but then one will hang the loop on the call to disable_monitor.

I gets the feels that I may have abused disable_monitor, or sockets, or both here.

Is there a right way to cleanup a socket and its monitor socket? I am willing to bet, when multiple sockets with monitors attached are concerned, I am probably not doing it right.

Jeducious:
OK, so, no need to wait, I decided to simply comment out the line that called disable_monitor and 'give it a ripper of a go' so tp speak. 

Now, the loop no longer hangs. So it seems calling disable_monitor is the wrong thing to do. Just don't know why. 

Should I:-
1. Just leave the disable_monitor commented out and live on in blissful ignorance now it apparently works
2. call close on the monitor socket instead of disable?
3. something else?

minrk:
If disable_monitor causes a hang, this suggests to me that there is a LINGER or ordering issue - that perhaps there are some messages not yet consumed by the monitor socket receiver, and the sender is blocking waiting for messages to be delivered.

That's a bit of a guess, though.

From [this discussion](https://github.com/zeromq/pyzmq/issues/1554#issuecomment-868457926) you need to call disable *before* close on the monitor socket (disable closes the socket that `bound`, which is handled internally by libzmq, while you need ot manage closing the socket that connects to listen for monitor messages)

Jeducious:
Thanks :) 

I eventually got the test suite to pass on macOS and windows using the following

```Python
        self.zmq_socket.disable_monitor()
        self.mon_sock.close(linger=0)
        self.zmq_socket.close(linger=0)
```

Which matches the discussion you just referred to, which I noticed I've actually been a part of. Seems that came back to bite me by not paying attention to it!


Confirming this now works fine on windows and macOS, and also linux now. 

I still have a run away condition of too many open fd's happening on linux. But that's a story for another day I think.

Issue #1526, opened 2021-05-13
async RCVTIMEO timeouts not cancelled upon successful recv


p4l1ly:
Currently, when using asyncio, the RCVTIMEO timeouts are handled in the following way. In addition to setting the RCVTIMEO option of the socket, a *timeout* asyncio future is created for each `recv` (or `recv_multipart`). The timeout future references the zmq `_Future` and cancels it after the timeout if the `_Future` is not `done()`. Now, imagine a client with a `REQ` socket that cyclically receives a big message from a `REP` socket and handles it shortly. The message cannot be garbage collected because it is still referenced by the timeout future until the specified RCVTIMEO timeout passes. If RCVTIMEO is big and the receivings are frequent, the dangling messages referenced by the dangling timeout futures may consume a lot of memory.

### Current Solution

https://github.com/zeromq/pyzmq/blob/145746d8e96c3c36882d5a5e85891de3b959a04f/zmq/_future.py#L292

### Evidence

```
# server.py
import asyncio
import zmq
import zmq.asyncio


async def srv():
    ctx = zmq.asyncio.Context()
    sock = ctx.socket(zmq.REP)
    sock.bind("tcp://127.0.0.3:5952")
    i = 0
    while True:
        print(i, flush=True)
        await sock.recv_multipart()
        await sock.send_multipart([b"." * 50000000])
        i += 1

asyncio.run(srv())
```

```
# client.py
import asyncio
import zmq
import zmq.asyncio
import gc
import linecache
import tracemalloc
import uvloop

uvloop.install()

def display_top(snapshot, key_type="lineno", limit=10):
    "return memory consumption log"
    snapshot = snapshot.filter_traces(
        (
            tracemalloc.Filter(False, "<frozen importlib._bootstrap>"),
            tracemalloc.Filter(False, "<unknown>"),
        )
    )
    top_stats = snapshot.statistics(key_type)

    ret = []
    ret.append("Top %s lines" % limit)
    for index, stat in enumerate(top_stats[:limit], 1):
        for line in stat.traceback.format():
            ret.append("#%s: %s:" % (index, line))
        frame = stat.traceback[0]
        ret.append(" " * 110 + "%.1f KiB" % (stat.size / 1024))
        line = linecache.getline(frame.filename, frame.lineno).strip()
        if line:
            ret.append("    %s" % line)

    other = top_stats[limit:]
    if other:
        size = sum(stat.size for stat in other)
[...]

PR #1527, opened 2021-05-13
_AsyncSocket: cancel timers upon success or exception

p4l1ly:
Resolves https://github.com/zeromq/pyzmq/issues/1526

### DIFF
--- a/zmq/_future.py
+++ b/zmq/_future.py
@@ -10,7 +10,7 @@ from typing import Type
 from zmq import EVENTS, POLLOUT, POLLIN
 import zmq as _zmq
 
-_FutureEvent = namedtuple('_FutureEvent', ('future', 'kind', 'kwargs', 'msg'))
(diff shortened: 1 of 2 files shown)


p4l1ly:
Both tornado's and asyncio's `call_later` returns an object that has the `cancel()` method. For other IOLoop implementations, `_call_later` can be overriden, returning e.g. `_NoTimer`.

minrk:
Thanks for the PR!


Looks like tornado timeouts don't always have cancel:

> AttributeError: '_Timeout' object has no attribute 'cancel'

This is in the "oldest supported tornado" test case, with tornado 4.5. We can bump minimum tornado to 5.x probably for `zmq.eventloop.future`, if that's when they introduced the cancel method. Or we can add an api shim.

p4l1ly:
`ioloop.remove_timeout` works in all versions of tornado, so I've decided to use it for tornado instead of `.cancel` which is compatible with asyncio but not compatible with tornado 4.5.

minrk:
Perfect, thanks!

PR #1535, opened 2021-05-25
try to cleanup contexts in test_decorators

minrk:
maybe leaks from here cause the occasional hang in test_devices on pypy

It's at least clear that the global context instance is leftover. However, the implication of the failure is that *sockets* are leftover, and I cannot see how that is happening.

it's the first test in test_devices that is flaky, and that test doesn't create any sockets, so it is likely leftover state from an earlier test and not the failing test itself that is the problem.


minrk:
FWIW, it doesn't make sense for this test to actually fix the issue since it doesn't close any additional *sockets*. Instead, it should only shift the hang earlier, to the test actually failing to do its cleanup. But 🤷 .

Issue #1548, opened 2021-06-14
Executables generated with pyinstaller do not import json correctly


dnltz:
Hey,

I wrote a Qt5 application which uses pyzmq. If I start the app with the Python interpreter, everything works and I can use the package. I just tried to create an executable from the app and run into the following problem:
```
Traceback (most recent call last):
  File "src/controller/connections.py", line 42, in keyPressEvent
    self.openMainWindow()
  File "src/controller/connections.py", line 37, in openMainWindow
    self.main = MainController(self.model.getData())
  File "src/controller/main.py", line 59, in __init__
    success, result = server.isConfigured()
  File "src/server.py", line 30, in isConfigured
    return self._send_command('status')
  File "src/server.py", line 54, in _send_command
    self.socket.send_json(command)
  File "zmq/sugar/socket.py", line 758, in send_json
  File "zmq/utils/jsonapi.py", line 41, in dumps
AttributeError: 'NoneType' object has no attribute 'dumps'
Fatal Python error: Aborted

Current thread 0x00007f5d7df99b80 (most recent call first):
  File "athena.py", line 11 in main
  File "athena.py", line 17 in <module>
Aborted (core dumped)
```
Seems like the jsonapi module is unable to import one of these three packages. Even install simplejson does not fix the issue. Here is a complete list of the packages:
```
altgraph==0.17
appdirs==1.4.3
CacheControl==0.12.6
certifi==2019.11.28
chardet==3.0.4
colorama==0.4.3
contextlib2==0.6.0
distlib==0.3.0
distro==1.4.0
html5lib==1.0.1
idna==2.8
importlib-metadata==4.5.0
ipaddr==2.2.0
lockfile==0.12.2
msgpack==0.6.2
packaging==20.3
pep517==0.8.2
progress==1.5
pyinstaller==4.3
pyinstaller-hooks-contrib==2021.1
pyparsing==2.4.6
PyQt5==5.14.2
PyQt5-sip==12.7.2
pytoml==0.1.21
PyYAML==5.3.1
pyzmq==22.1.0
requests==2.22.0
retrying==1.3.3
six==1.14.0
typing-extensions==3.10.0.0
urllib3==1.25.8
webencodings==0.5.1
zipp==3.4.1
```

BTW: The executable was generated in a virtual environment using py37 on an Ubuntu 20.04.

[...]

PR #1549, opened 2021-06-17
remove optional JSON imports from jsonapi

minrk:
It's been a long time since json wasn't in the standard library. Let's get rid of this complexity and let folks use their own JSON implementation via custom serializers.

Avoids issues like #1548

### DIFF
--- a/zmq/eventloop/zmqstream.py
+++ b/zmq/eventloop/zmqstream.py
@@ -280,11 +280,8 @@ class ZMQStream(object):
         """Send json-serialized version of an object.
         See zmq.socket.send_json for details.
         """
-        if jsonapi is None:
-            raise ImportError('jsonlib{1,2}, json or simplejson library is required.')
-        else:
-            msg = jsonapi.dumps(obj)
-            return self.send(msg, flags=flags, callback=callback, **kwargs)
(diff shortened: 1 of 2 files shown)


martinRenou:
I was just looking at this code, and I was wondering if some other JSON libraries could be added. It might be interesting to use `rapidjson` or `orjson` if available, as they are quite more efficient than the standard lib.

Happy to open an issue or try a PR if you agree.

minrk:
That's a great idea, but I don't think it belongs in pyzmq. An *example* would be appropriate, though. In the context of Jupyter, doing it at the  level of `jupyter_client.session` would make more sense. Anyone who cares about serialization performance ought to be using custom serialization and only `socket.send_multipart()` (like Jupyter does).

martinRenou:
> An example would be appropriate, though

Yeah sorry, I was deep into the jupyter code and only had that in mind :P 

> doing it at the level of jupyter_client.session would make more sense. Anyone who cares about serialization performance ought to be using custom serialization and only socket.send_multipart() (like Jupyter does).

Makes sense. I could try a PR on the jupyter_client side then. Maybe I will open an issue first.

Issue #1555, opened 2021-07-02
FATAL ERROR: OUT OF MEMORY 


RichardLions:
Hello

I am trying to debug a python load test package called Locust.

It creates a ROUTER and several DEALERs. The ROUTER often crashes due to this assert: warning : FATAL ERROR: OUT OF MEMORY (C:\projects\libzmq\src\decoder_allocators.cpp:85)

It is setup to use zmq.green.

This is running on a windows machine using localhost.

pyzmq version 22.1.0

I doubt this is something wrong with pyzmq. What sort of misuse could cause such an error to be thrown?

Thanks

Matthew--Townsend:
I have the same issue with Locust and version 22.1.0 of pyzmq.  I am new to using it so I don't have any idea if other versions have worked.

python 3.9.5
locust 1.5.3
pyzmq 22.1.0

FATAL ERROR: OUT OF MEMORY (C:\projects\libzmq\src\decoder_allocators.cpp:85).

minrk:
The call that's failing is [malloc](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/malloc). I'm not an expert in Windows memory management. But possible causes include:

- allocating a very large message (larger than _HEAP_MAXREQ), which is ~ 4 GB when run under win32
- failure to deallocate messages (they could be queued in memory if they cannot be delivered, or they might not be freed properly)

Unfortunately, it's hard to say much without reproducible test cases.

RichardLions:
Hello

I believe the issue is std::size_t shared_message_memory_allocator::_max_counters is being overridden by mistake from somewhere.

I wrote some debug to output the values of _max_counters when calls to shared_message_memory_allocator are made. The normal value should be 249 however it changes to large values such as 1947977969464. The value it changes to is never the same each test but is always very large. The assert does not trigger when _max_counters changes. The value can be in a malformed state for most of the applications lifetime. If an allocation is made it will use the large _max_counters which triggers the assert due to trying to allocate an incredibly large amount of memory.

Note: _max_counters can actually be made const. When doing so the value stops become malformed but I suspect this is the compiler hiding the memory override somehow.

I have attached a small folder for VSCode that will reproduce the issue. It is about 1/5 repro rate for the assert. However most runs will malform _max_counters.

[libzmq_repro.zip](https://github.com/zeromq/pyzmq/files/6771022/libzmq_repro.zip)

minrk:
[...]

PR #1560, opened 2021-07-09
Updated core.py to use close() instead of cancel() on Timeout objects

RichardLions:
Timeout.close() should be used when the Timeout object will no longer be used.
https://www.gevent.org/api/gevent.timeout.html#gevent.Timeout

This fixes issue: https://github.com/zeromq/pyzmq/issues/1555

Creating to many Timeout objects without closing them causes memory to be overwritten. This was causing libzmq to assert. I do not believe this should be the correct behaviour for Timeout. I am going to open an issue on the gevent repo to inform them what I have found out.

I used gflags to enable heap debugging on python.exe to thrown an exception when memory was overridden to find this issue.
https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/gflags-commands

### DIFF
--- a/zmq/green/core.py
+++ b/zmq/green/core.py
@@ -161,7 +161,7 @@ class _Socket(_original_Socket):
                 )
         finally:
             if timeout:
-                timeout.cancel()
+                timeout.close()
             self.__writable.set()
 
     def _wait_read(self):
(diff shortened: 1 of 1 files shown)


Matthew--Townsend (commented):
I have implemented this change locally on my box in C:\Program Files\Python39\Lib\site-packages\zmq\green\core.py.  After doing so, the out of memory issues that I was getting with locust (uses pyzmq) are gone.  Thanks for finding and fixing!

minrk:
Great find, thank you!

Issue #1561, opened 2021-07-14
Latest pyzmq versions fail to pass Apple notarization


TechnicalPirate:
Hello,

**Problem**: We convert our python tool into a binary using `cx_freeze` and pass it to Apple for notarization prior to deployment to client machines. When using Python 3.9 and `pyzmq-22.1.0` - pyzmq fails to pass [Apples notarization policies.](https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution)

Looking at the pyzmq changelog i guess this would have started to manifest on [version 21](https://pyzmq.readthedocs.io/en/latest/changelog.html#id7)? 

**Notarization Error Report**:

```
{
    "severity": "error",
    "code": null,
    "path": "OurApp.app/Contents/Resources/lib/zmq/.dylibs/libsodium.23.dylib",
    "message": "The binary is not signed.",
    "docUrl": null,
    "architecture": "x86_64"
},

{
    "severity": "error",
    "code": null,
    "path": "OurApp.app/Contents/Resources/lib/zmq/.dylibs/libsodium.23.dylib",
    "message": "The signature does not include a secure timestamp.",
    "docUrl": null,
    "architecture": "x86_64"
},

{
    "severity": "error",
    "code": null,
    "path": "OurApp.app/Contents/Resources/lib/zmq/.dylibs/libsodium.23.dylib",
    "message": "The binary is not signed with a valid Developer ID certificate.",
    "docUrl": null,
    "architecture": "arm64"
},

{
    "severity": "error",
    "code": null,
    "path": "OurApp.app/Contents/Resources/lib/zmq/.dylibs/libsodium.23.dylib",
    "message": "The signature does not include a secure timestamp.",
    "docUrl": null,
    "architecture": "arm64"
},

{
    "severity": "error",
    "code": null,
    "path": "OurApp.app/Contents/Resources/lib/zmq/.dylibs/libzmq.5.dylib",
    "message": "The binary is not signed.",
    "docUrl": null,
    "architecture": "x86_64"
},

{
    "severity": "error",
    "code": null,
    "path": "OurApp.app/Contents/Resources/lib/zmq/.dylibs/libzmq.5.dylib",
    "message": "The signature does not include a secure timestamp.",
    "docUrl": null,
    "architecture": "x86_64"
},

{
    "severity": "error",
    "code": null,
    "path": "OurApp.app/Contents/Resources/lib/zmq/.dylibs/libzmq.5.dylib",
[...]

PR #1565, opened 2021-07-24
bump cibuildwheel to 2.0

minrk:
this bumps delocate from 0.8 to 0.9, which includes adhoc code signing, which may help with freezing applications (#1561)

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -157,16 +157,16 @@ jobs:
               build: "cp*win32"
 
           - os: windows-2016
-            name: win32-pypy
-            architecture: x86
+            name: win-pypy
+            architecture: x64
             cibw:
-              build: "pp*win32"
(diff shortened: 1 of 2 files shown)


minrk:
Oops, delocate isn't bumped to 0.9 until https://github.com/pypa/cibuildwheel/pull/771 is merged and released

.github/workflows/wheels.yml
```diff
@@ -212,16 +212,6 @@ jobs:
           pip install --upgrade setuptools pip wheel
           pip install cibuildwheel=="${{ env.cibuildwheel }}" cython=="${{ env.cython }}"
 
-      - name: install mac dependencies
```
minrk, line 215: These were redundant - cibuildwheel ignores the host environment and uses its own pinned install, controlled via $CIBW_DEPENDENCY_VERSIONS

Issue #1566, opened 2021-07-26
How to create a mutable numpy array from a non-copying recv?


ronawho:
[serialsocket.py](https://github.com/zeromq/pyzmq/blob/c02e8a1094be8d817020af221a283176ff22eed5/examples/serialization/serialsocket.py) has an example of how to recv a numpy array, but this doesn't seem to work with the cffi backend. Running this example or [test_buffer_numpy](https://github.com/zeromq/pyzmq/blob/f0b60e2e9cfc0d51b4db0f24c7d795bd94d79c1f/zmq/tests/test_message.py#L306-L335) under PyPy I see errors like:

```python
Traceback (most recent call last):
  File "examples/serialization/serialsocket.py", line 78, in <module>
    main()
  File "examples/serialization/serialsocket.py", line 70, in main
    C = req.recv_array(copy=False)
  File "examples/serialization/serialsocket.py", line 47, in recv_array
    A = numpy.frombuffer(msg, dtype=md['dtype'])
TypeError: 'Frame' does not support the buffer interface
```

[serialization.rst](https://github.com/zeromq/pyzmq/blob/c02e8a1094be8d817020af221a283176ff22eed5/docs/source/serialization.rst) has a slightly different example that explicitly wraps the `Frame` with a `memoryview`, but that fails in a similar manner under PyPy. Wrapping with a `memoryview` also makes the numpy array readonly with the cython backend.

Using `Frame.buffer` seems to work for both backends, but it also seems to create a readonly array with the cython backend. e.g.:

```diff
diff --git a/examples/serialization/serialsocket.py b/examples/serialization/serialsocket.py
index 824f02b..5a5e766 100644
--- a/examples/serialization/serialsocket.py
+++ b/examples/serialization/serialsocket.py
@@ -43,7 +43,7 @@ class SerializingSocket(zmq.Socket):
     def recv_array(self, flags=0, copy=True, track=False):
         """recv a numpy array"""
         md = self.recv_json(flags=flags)
-        msg = self.recv(flags=flags, copy=copy, track=track)
+        msg = self.recv(flags=flags, copy=copy, track=track).buffer
         A = numpy.frombuffer(msg, dtype=md['dtype'])
         return A.reshape(md['shape'])

@@ -72,6 +72,7 @@ def main():
     print("Okay" if (A == B).all() else "Failed")
     print("Checking send_array...")
     print("Okay" if (C == B).all() else "Failed")
[...]

PR #1567, opened 2021-07-27
Fix memoryview support for the cffi backend

ronawho:
Previously, `__buffer__` was returning the raw `_buffer`, which isn't
always set. This resulted in `None` being returned, which caused type
errors about Frame not supporting the buffer interface. Switch to
returning `buffer`, which will set `_buffer` if it isn't already.

Part of #1566


minrk:
Thanks! Affected tests pass. The timeouts are due to pip's very slow dependency solver, which I'm fixing in #1565

PR #1570, opened 2021-07-30
Added support for Windows IPC in bundled libzmq

### LINKED ISSUE #1462: IPC on Windows still unavailable in v20.0
*Please use this template for reporting suspected bugs or requests for help.*

# Issue description

With pyzmq 20.0 linking against libzmq 4.3.3, `icp://` is expected to work in Windows 10 when `AF_UNIX` is available. However,  `zmq.error.ZMQError: Protocol not supported` is raised.

# Environment

* pyzmq 20.0.0
* libzmq 4.3.3
* python 3.8.6 (and possibly others)
* OS: Windows 10, Version10.0.19041 Build 19041

# Minimal test code / Steps to reproduce the issue

Ensure AF_UNIX is available `sc query afunix` then

```python
import zmq
context = zmq.Context()
socket = context.socket(zmq.PUB)
socket.bind('ipc:///tmp/stream')
```

# What's the actual result? (include assertion message & call stack if applicable)
`zmq.error.ZMQError: Protocol not supported`

# What's the expected result?
bound socket.

mattrjackson:
A few small changes to enable IPC under Windows (resolves #1462). This does create an external dependency on wepoll, but this is already a dependency for libzmq.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "externals/wepoll"]
+	path = externals/wepoll
+	url = https://github.com/piscisaureus/wepoll.git
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,2 @@
 docs/source/_templates
+externals
--- a/buildutils/check_sys_un.c
+++ b/buildutils/check_sys_un.c
@@ -1,5 +1,9 @@
 #include <stdio.h>
-#include "sys/un.h"
+#if defined _MSC_VER
+#include <afunix.h>
+#else
+#include <sys/un.h>
+#endif
 
 int main(int argc, char **argv) {
     struct sockaddr_un *dummy;
--- a/buildutils/include_win32/platform.hpp
+++ b/buildutils/include_win32/platform.hpp
@@ -17,8 +17,10 @@
 
 // additional defines required for 4.3 after this file was removed
 #define ZMQ_USE_SELECT 1
-#define ZMQ_IOTHREAD_POLLER_USE_SELECT 1
 #define ZMQ_POLL_BASED_ON_SELECT 1
 #define ZMQ_USE_CV_IMPL_WIN32API 1
+#define ZMQ_IOTHREAD_POLLER_USE_EPOLL 1
 
+/* Have AF_UNIX sockets for ipc transport */
+#define ZMQ_HAVE_IPC 1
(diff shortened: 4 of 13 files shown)


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "externals/wepoll"]
```
minrk, line 1: Is there a way to do this without submodules? I'd really like to avoid bringing in a submodule dependency here, as they've caused me a lot of problems in the past (e.g. `git archive` doesn't work anymore, so GitHub download links don't work).

Looking at how [tiny wepoll is](https://github.com/piscisaureus/wepoll), maybe vendoring it with [git subtree](https://www.atlassian.com/git/tutorials/git-subtree) is the way to go.

minrk (commented):
Wonderful, thank you for the PR!

.gitmodules
mattrjackson: works for me.  just updated accordingly.

minrk:
Great! I rebased it to use a squashed `git-subtree` instead. You can run the autoformatter with `pre-commit run --all-files` to fix the lint issues.

There are some test failures:

1. the `zmq.has` test is failing because it currently assumes Windows won't have ipc, that test should be updated
2. build failures on non-Windows. The `wepoll.c` should probably be added to sources only on Windows, so it doesn't try to compile wepoll on all platforms.


I should also note that this should not have any effect on pyzmq wheels, which do not use pyzmq-built libzmq. They use [these public releases from libzmq](https://github.com/zeromq/libzmq/releases/). If those don't have ipc support, I think that's an issue to bring up with libzmq.

mattrjackson:
Fixed those issues. I think the standard libzmq library will work fine (I'm currently using what I think is a stock 4.3.4 dll with JNI), but if not, I can work on that with the libzmq team.

minrk:
> I can update bundle.py to fix this if you want, but it will mean using 141 for x86 and 142 for x64.

That would be great! If you open a PR updating bundle.py, it will create new wheels that you can download and test from the actions > artifacts link on the PR.

AntiTenzor:
Thank you very much!
I really expect IPC to be much faster, than TCP.
I was really upset, that there was no fast and effective IPC support for such beautiful gem as ZeroMQ (in Windows universe).

PR #1572, opened 2021-07-30
double check symlinks in generate_certificates

andrew-azarov:


minrk:
Have you tried this? `shutil.rmtree` cannot be run on a symlink, it raises:

```python
OSError: Cannot call rmtree on a symbolic link
```

andrew-azarov:
It will safely remove the folder if it is mounted though. So the symlink protection will save it from a symlink but not from hardlink (default in zfs for example) or mounting (ouch).

PR #1574, opened 2021-08-03
Updating bundle.py for x64 Windows

mattrjackson:
This updates bundle.py to use v142 of the libzmq dll, which appears necessary to support Windows IPC.


mattrjackson:
Just tested the x64 wheel, it works on my win10 system.

minrk:
Wonderful, thank you for confirming!

PR #1584, opened 2021-08-09
Add readthedocs document link to readme

sharteeya:
Hi!
Just found that the document on the Github page is disabled, but the one on Read the Docs is still activated. I think it's good to add another link to the document so users can still access the document if one of them is offline.

Thank you for keeping this project great!


minrk:
Thanks!

PR #1586, opened 2021-08-09
fix: fixed pubsub example py2 syntax

j8xixo12:
Hi, 
    When I executed the `pubsub` example, I encountered some syntax error, and I found these syntax error coming from some `python2` syntax in example code of `pubsub`.

I fixed these `py2` syntax in `pubsub` !

### DIFF
--- a/examples/pubsub/subscriber.py
+++ b/examples/pubsub/subscriber.py
@@ -26,7 +26,7 @@ def sync(connect_to):
     ctx = zmq.Context.instance()
     s = ctx.socket(zmq.REQ)
     s.connect(sync_with)
-    s.send('READY'.encode('utf-8'))
+    s.send(b'READY')
     s.recv()
 
 
@@ -45,7 +45,7 @@ def main():
     ctx = zmq.Context()
     s = ctx.socket(zmq.SUB)
     s.connect(connect_to)
-    s.setsockopt(zmq.SUBSCRIBE, ''.encode('utf-8'))
+    s.setsockopt(zmq.SUBSCRIBE, b'')
 
     sync(connect_to)
 
--- a/examples/pubsub/subscriber.py
+++ b/examples/pubsub/subscriber.py
@@ -26,7 +26,7 @@ def sync(connect_to):
     ctx = zmq.Context.instance()
     s = ctx.socket(zmq.REQ)
     s.connect(sync_with)
-    s.send('READY')
+    s.send('READY'.encode('utf-8'))
     s.recv()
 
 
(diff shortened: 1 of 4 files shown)


examples/pubsub/subscriber.py
```diff
@@ -45,23 +45,23 @@ def main():
     ctx = zmq.Context()
     s = ctx.socket(zmq.SUB)
     s.connect(connect_to)
-    s.setsockopt(zmq.SUBSCRIBE, '')
+    s.setsockopt(zmq.SUBSCRIBE, ''.encode('utf-8'))
 
     sync(connect_to)
 
-    start = time.clock()
+    start = time.process_time()
```
j8xixo12, line 52: `clock()` was removed since `python 3.3` and replaced it with `process_time()` and `perf_counter()`

examples/pubsub/subscriber.py
```diff
@@ -45,23 +45,23 @@ def main():
     ctx = zmq.Context()
     s = ctx.socket(zmq.SUB)
     s.connect(connect_to)
-    s.setsockopt(zmq.SUBSCRIBE, '')
+    s.setsockopt(zmq.SUBSCRIBE, ''.encode('utf-8'))
```
mlouielu, line 48: `b''` may be simple at this case

examples/pubsub/topics_pub.py
```diff
@@ -55,9 +55,9 @@ def main():
     msg_counter = itertools.count()
     try:
         for topic in itertools.cycle(all_topics):
-            msg_body = str(msg_counter.next())
+            msg_body = str(next(msg_counter))
```
mlouielu, line 58: We can do byte encoding at here.

examples/pubsub/topics_pub.py
```diff
@@ -55,9 +55,9 @@ def main():
...
             print('   Topic: %s, msg:%s' % (topic, msg_body))
-            s.send_multipart([topic, msg_body])
+            s.send_multipart([topic.encode('utf-8'), msg_body.encode('utf-8')])
```
mlouielu, line 60: Maybe update topic definition? so we don't need to encode topic repeatedly.

minrk (commented):
Thanks for the PR and thanks @mlouielu for the review! I agree with all the points made so far.

j8xixo12:
@mlouielu thanks for the review !!

mlouielu:
LGTM.

But as pyzmq support py2 & py3, do we need to provide both version of example for them @minrk ?

minrk:
Thank you both!

mlouielu:
> > But as pyzmq support py2 & py3,
> 
> latest versions of pyzmq require Python >= 3.6, so we can use f-strings and everything.

Nice to heard that!

PR #1592, opened 2021-08-11
ResourceWarnings

minrk:
add resource warnings to track down unclosed sockets/contexts:

Example:

```python
import tracemalloc
import zmq
import warnings


def f():
    ctx = zmq.Context()
    s = ctx.socket(zmq.PUSH)

def main():
    f()

if __name__ == "__main__":
    tracemalloc.start(2)
    warnings.simplefilter('default')
    main()
```

gives:

```python
testresource.py:12: ResourceWarning: unclosed socket <zmq.Socket(zmq.PUSH) at 0x109be3160>
  f()
Object allocated at (most recent call last):
  File "testresource.py", lineno 8
    s = ctx.socket(zmq.PUSH)
  File "dev/zmq/pyzmq/zmq/sugar/context.py", lineno 263
    s = self._socket_class(  # set PYTHONTRACEMALLOC=2 to get the calling frame
testresource.py:12: ResourceWarning: unclosed context <zmq.Context() at 0x1096590e0>
  f()
Object allocated at (most recent call last):
  File "testresource.py", lineno 12
    f()
  File "testresource.py", lineno 7
    ctx = zmq.Context()
```

### DIFF
--- a/zmq/sugar/context.py
+++ b/zmq/sugar/context.py
@@ -63,9 +63,10 @@ class Context(ContextBase, AttributeSetter):
 
         if not self._shadow and not _exiting and not self.closed:
             warnings.warn(
-                f"Implicitly terminating {self}. Make sure to call Context.term().",
+                f"unclosed context {self}",
                 ResourceWarning,
                 stacklevel=2,
+                source=self,
             )
             self.term()
 
@@ -259,7 +260,9 @@ class Context(ContextBase, AttributeSetter):
         """
         if self.closed:
             raise ZMQError(ENOTSUP)
-        s = self._socket_class(self, socket_type, **kwargs)
+        s = self._socket_class(  # set PYTHONTRACEMALLOC=2 to get the calling frame
+            self, socket_type, **kwargs
+        )
         for opt, value in self.sockopts.items():
             try:
                 s.setsockopt(opt, value)
--- a/zmq/sugar/context.py
+++ b/zmq/sugar/context.py
@@ -9,6 +9,7 @@ import os
 from threading import Lock
 from typing import TypeVar, Type
 from weakref import WeakSet
+import warnings
 
 from zmq.backend import Context as ContextBase
 from . import constants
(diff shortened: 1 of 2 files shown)


davetapley:
@minrk I just started seeing `ResourceWarning: unclosed socket` show up (which led me to this PR) but now I'm confused:

My assumption and [these docs](https://pyzmq.readthedocs.io/en/latest/api/zmq.html#zmq.Socket.close) say:

> the socket will automatically be closed when it is garbage collected.

---

But re-reading those docs I see:

> Note: closing a zmq Socket may not close the underlying sockets if there are undelivered messages. Only after all messages are delivered or discarded by reaching the socket’s LINGER timeout (default: forever) will the underlying sockets be closed.

To be sure:
Is the '`socket`' in the `ResourceWarning: unclosed socket` warning the 'underlying socket', or the `zmq.Socket`? 🤔

minrk:
> Is the 'socket' in the ResourceWarning: unclosed socket warning the 'underlying socket', or the zmq.Socket? 🤔

the zmq.Socket. pyzmq doesn't actually know anything about the underlying socket, that's all internal to libzmq.

Just like other standard library sockets and files that show resource warnings when they are closed by garbage collection, the ResourceWarning is shown to indicate the situation: pyzmq is closing a socket for you, but it's a good idea to close it yourself.

For example:

```python
def main():
    f = open("/tmp/f", "w")

if __name__ == "__main__":
    main()
```

If you run that with `PYTHONWARNINGS=d`, you'll get:

```
test.py:6: ResourceWarning: unclosed file <_io.TextIOWrapper name='/tmp/f' mode='w' encoding='UTF-8'>
```

The file _is_ closed, but you get a warning because the close was _implicit_ instead of explicit. pyzmq sockets now have similar behavior to standard library files and sockets in this regard. Relying on garbage collection is not always safe in Python, so it's a good idea to close things explicitly.

#1845 now mentions the resource warnings that you'll see if you rely on garbage collection

PR #1601, opened 2021-08-22
Get libzmq from OBS builds for arm wheels

minrk:
aarch64 builds of libzmq are too expensive to do in our own CI, so we use binaries. Previously, the only binaries I could find for manylinux2014 (CentOS 7) were pretty out of date, but we can get them from upstream OBS builds<del> for debian 9 (still no builds for CentOS 7 + aarch64)</del>.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -205,7 +205,8 @@ jobs:
         # it's not in default packages, but it is in epel-7
         # this will be an older release than than other wheels
         run: |
-          echo 'CIBW_BEFORE_ALL_LINUX=yum -y install epel-release && yum -y install zeromq-devel' >> "$GITHUB_ENV"
+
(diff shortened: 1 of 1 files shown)


bluca:
I've enabled CentOS 7 aarch64 builds, should be available in a few minutes

minrk:
@bluca nice! I've struggled to find docs on how to install things from OBS. Do you have a sample or link for installing from an OBS repo for CentOS?

bluca:
The "download" button on each package (eg: top right at https://build.opensuse.org/package/show/network:messaging:zeromq:git-stable/libzmq ) is supposed to give you instructions, but unfortunately it matches the wrong thing (assumes source package name == binary package name).
Adjusting the URL it gives a little to match the binary package then gives you the information, eg:

https://software.opensuse.org//download.html?project=network%3Amessaging%3Azeromq%3Agit-stable&package=libzmq5

minrk:
exactly what I was looking for, thanks

PR #1615, opened 2021-09-21
OBS CI: use link_package instead of branch_package

bluca:
It should keep the same settings as the source, thus skipping disabled builds


bluca:
(this will have effect only on the next PR)

bluca:
This doesn't cut it either, disabled again

PR #1624, opened 2021-10-10
fix keys directory

mazaheriaan:
The keys are not in the certificate directory


minrk:
I think the directories are doing what they mean to (there's the original `certificates` directory, whose contents are populated by the `generate_certificates` utility, but that is reorganized into sibling private/public key dirs).

If anything's missing, it's perhaps the removal of the temporary `certificates` directory, which is empty after running `generate_certificates.py`. What do you think?

minrk:
If you want to update the PR to remove the empty dir instead, I'd be happy to merge.

PR #1631, opened 2021-11-05
add a few more types

### LINKED ISSUE #1627: Mypy thinks Context.socket is an untyped function
Consider this minimal example:

```python
import zmq

ctx = zmq.Context.instance()
socket = ctx.socket(zmq.PULL)
```
When running mypy with the  `disallow_untyped_calls` option, an error surfaces about this function being untyped:

```
$ mypy --disallow-untyped-calls example.py 
example.py:4: error: Call to untyped function "socket" in typed context  [no-untyped-call]
```

My workaround is to annotate directly and ignore the error:

```
from zmq.sugar.socket import Socket
socket: Socket = ctx.socket(zmq.PULL)  # type: ignore
```

I've looked at https://github.com/zeromq/pyzmq/issues/1206 and it doesn't seem like [`context.py`](https://github.com/zeromq/pyzmq/blob/main/zmq/sugar/context.py) has either inline annotations or an associated `.pyi` file. Should this work out of the box?

minrk:
in particular, Context.term and Context.socket

closes #1627


minrk:
Not everything is typed, but the key bits are

PR #1633, opened 2021-11-05
add isort, pyupgrade to pre-commit

minrk:


minrk:
will rebase after finishing #1617

PR #1634, opened 2021-11-06
Make docstring match type annotation

dmwyatt:
I *think* the type annotation of `Optional[int]` is correct and the docstring of `float, int` is incorrect.  

The generated API docs use the docstring and so indicate that you can use a float or an int, but then when you use a float, you get type errors in your IDE/mypy/etc.


minrk:
Technically float is accepted, but immediately cast to int so truncated. This seems right to me.

Issue #1646, opened 2022-01-08
ROUTER-ROUTER communication example


jcpinto54:
Here is an example of a ROUTER-ROUTER communication for anyone who is building a P2P application and needs a socket that is able to connect to several other peers simultaneously, while also being able to receive messages. I'm posting this here because I didn't find an example anywhere and it might be useful for someone.

The program takes two arguments: 
1. "s" or "c", meaning server or client.
2. If the first argument is "c", then this argument is used as the identity of the peer. Can be any string.

Note that starting the program as a server sets the peer identity in the connection as "server".

```py
import zmq
import sys

context = zmq.Context()

socket = context.socket(zmq.ROUTER)

if sys.argv[1] == "s":
    socket.setsockopt_string(zmq.IDENTITY, 'server')
    socket.bind("tcp://*:4000")

    while True:
        print("waiting for msg")
        req = socket.recv_multipart()
        print(req)
        #                      identity of receptionist   empty frame     message content
        socket.send_multipart([req[0],                    bytes(),      bytes("whatup", encoding="utf-8")])

elif sys.argv[1] == "c":
    socket.setsockopt_string(zmq.IDENTITY, sys.argv[2])
    socket.connect("tcp://localhost:4000")

    while True:
        input("enter to send msg")
        #                       identity of receptionist          empty frame     message content
        socket.send_multipart([bytes("server", encoding="utf-8"), bytes(),     bytes("cheers", encoding="utf-8")])

        rep = socket.recv_multipart()
        print(rep)

# In real applications do not forget to disconnect when the connection is not needed anymore

```

frobnitzem:
Well, I've been spoiling to put in an example of aiowire, so I've created PR #1652 to add this example into the asyncio examples subdirectory.  From your comments above, I assume you intended to use CC0 (public domain copyright).

mrvn:
How would this example look with CURVE encryption?

PR #1652, opened 2022-02-04
Added router_router example.

frobnitzem:
This PR adds the example created in issue #1646 , then runs 3 ROUTERS at once using asyncio.

### DIFF
--- a/examples/asyncio/router_router.py
+++ b/examples/asyncio/router_router.py
@@ -12,16 +12,18 @@ Contributed by github:jcpinto54 and github:frobnitzem
 
 import asyncio
 
+from aiowire import EventLoop
+
 import zmq
 from zmq.asyncio import Context
-from aiowire import EventLoop
+
 
 class Server:
     def __init__(self, url):
         context = Context.instance()
         socket = context.socket(zmq.ROUTER)
         socket.setsockopt_string(zmq.IDENTITY, 'server')
-        socket.bind( url )
+        socket.bind(url)
         self.socket = socket
 
     async def run(self, ev):
@@ -35,6 +37,7 @@ class Server:
         # continue to run by returning this `Wire`
         return self.run
 
+
 class Client:
     def __init__(self, url, name):
         self.name = name
@@ -42,7 +45,7 @@ class Client:
         context = Context.instance()
         socket = context.socket(zmq.ROUTER)
         socket.setsockopt_string(zmq.IDENTITY, name)
-        socket.connect( url )
+        socket.connect(url)
         self.socket = socket
 
     async def run(self, ev):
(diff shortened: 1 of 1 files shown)


examples/asyncio/router_router.py
```diff
@@ -0,0 +1,75 @@
...
+        await self.socket.send_multipart([b'server', b'', b'cheers'])
+        rep = await self.socket.recv_multipart()
+        print(f"{self.name} received {rep}")
+
+        # continue to run by returning this `Wire`
+        return self.run
+
+
+url = 'inproc://test_zmq'
+srv = Server(url)
```
minrk, line 63: Given the pending deprecation of globals like `asyncio.get_event_loop()`, I'd like to encourage creating asyncio objects within the `async def main`

examples/asyncio/router_router.py
```diff
@@ -0,0 +1,75 @@
...
+romeo = Client(url, 'romeo')
+sierra = Client(url, 'sierra')
+
+
+async def main():
+    async with EventLoop(1.0) as ev:
```
minrk, line 69: What does this EventLoop accomplish? It seems these all should work with stdlib `asyncio.create_task(romeo.run())`, etc.

minrk (commented):
Thanks for the example! I'm not quite clear on what aiowire's doing here, but maybe just a little clarification in the comments of what it's adding to stdlib asyncio is all that's needed.

examples/asyncio/router_router.py
frobnitzem: The `EventLoop` is awaiting all tasks launched within it, while timing out after (here) 1 second.  This guarantees completion of tasks - even if they are infinite loops - as long as they regularly call await.

examples/asyncio/router_router.py
frobnitzem: I don't _think_ that any of the parts of the `Client` or `Server` `__init__` functions call `get_event_loop` - so Client and Server classes are not themselves asyncio objects - unless zmq.asyncio.Context is an asyncio object.

I guess the worry is that some code outside of an async function might call Socket.recv or Server.run, which would be incorrect.

examples/asyncio/router_router.py
minrk: Cool! I think that's worth explaining what aiowire's responsible for in the example, since it's not clear from just the code itself what's happening.

examples/asyncio/router_router.py
minrk: That's true for pyzmq >= 22.2, but pyzmq 22.1 called `get_event_loop` in . I think it's still a pattern to encourage: do not create module-global async objects, even if creating the objects themselves _may_ not yet grab a handle on the loop.

frobnitzem:
I've addressed issues raised in PR comments.

minrk:
Thanks!

PR #1663, opened 2022-03-02
Fix doc typos upd -> udp

wimglenn:
upd visible for example on https://pyzmq.readthedocs.io/en/latest/api/zmq.html


minrk:
Thanks!

PR #1673, opened 2022-03-17
Update comment in poll.py

danielg1111:
I changed a comment, since the conversion described appears to be to *milliseconds* rather than *microseconds*


minrk:
Thanks!

PR #1684, opened 2022-04-07
Add missing return types.

oxidase:
The PR fixes MYPY errors:
```
../zmq/backend/__init__.pyi:52: error:
Function is missing a return type annotation  [no-untyped-def]
    def zmq_poll(sockets: List[Any], timeout: Optional[int] = ...): ...
    ^
../zmq/backend/__init__.pyi:53: error:
Function is missing a return type annotation  [no-untyped-def]
    def device(device_type: int, frontend: Socket, backend: Optional[Socke...
    ^
../zmq/backend/__init__.pyi:54: error:
Function is missing a return type annotation  [no-untyped-def]
    def proxy(frontend: Socket, backend: Socket): ...
    ^
../zmq/backend/__init__.pyi:55: error:
Function is missing a return type annotation  [no-untyped-def]
    def proxy_steerable(
    ^
Found 4 errors in 1 file (checked 2 source files)
```


minrk:
Thanks!

PR #1688, opened 2022-04-25
Hold a reference to `warnings.warn` for use in __del__

chrisjbillington:
I've been getting output like this when Python exits, similar to https://github.com/zeromq/pyzmq/issues/1356.

```
Exception ignored in: <function Socket.__del__ at 0x7f02b08ebf70>
  File "/usr/lib/python3.8/site-packages/zmq/sugar/socket.py", line 97, in __del__
TypeError: 'NoneType' object is not callable
```
Holding a reference to `warnings.warn`, similar to the fix for #1356, seems to resolve the problem.

### DIFF
--- a/zmq/sugar/socket.py
+++ b/zmq/sugar/socket.py
@@ -8,7 +8,7 @@ import errno
 import pickle
 import random
 import sys
-import warnings
+from warnings import warn
 from typing import (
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

PR #1702, opened 2022-05-23
was impossible to do zmq.utils.monitor.recv_monitor_message on an AsyncIO socket

shimondoodkin:
the method recv_monitor_message was trying to parse a asyncio.Future 

here is a fix

### DIFF
--- a/zmq/utils/monitor.py
+++ b/zmq/utils/monitor.py
@@ -105,6 +105,8 @@ def recv_monitor_message(socket: zmq.Socket, flags: int = 0) -> Union[_MonitorMe
     -------
     event : dict
         event description as dict with the keys `event`, `value`, and `endpoint`.
+
+    if socket.context is of zmq.asyncio.Context type, return asyncio.Future
     """
     
     #transparently handle asyncio socket, jsut return a future instead of a dict
--- a/zmq/utils/monitor.py
+++ b/zmq/utils/monitor.py
@@ -19,6 +19,11 @@ class _MonitorMessage(TypedDict):
 
 def parse_monitor_message(msg: List[bytes]) -> _MonitorMessage:
     """decode zmq_monitor event messages.
+    
+    If you are getting error of:
+        'TypeError: object of type '_asyncio.Future' has no len()'
+        'RuntimeError: There is no current event loop in thread 'Thread-1'
+    then use the async monitor
 
     Parameters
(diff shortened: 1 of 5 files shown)


zmq/utils/monitor.py
```diff
@@ -70,7 +107,15 @@ def recv_monitor_message(socket: zmq.Socket, flags: int = 0) -> _MonitorMessage:
     -------
     event : dict
         event description as dict with the keys `event`, `value`, and `endpoint`.
+
+    if socket.context is of zmq.asyncio.Context type, return asyncio.Future
     """
+
+    # transparently handle asyncio socket,
+    # returns a future instead of a dict
+    if isinstance(socket.context, zmq.asyncio.Context):
```
minrk, line 116: I don't think this is the way to check for an async socket. Instead, call `msg = socket.recv_multipart` and then check if `msg` is awaitable. If so, wrap the future.

shimondoodkin: i don't know to do it,  please help

shimondoodkin: do you mean like to use 
https://docs.python.org/3/library/inspect.html#inspect.isawaitable
if it is awaitable then , to call the other async function and pass the awaitable variable as an argument, and await it there?

shimondoodkin: like this?
https://github.com/shimondoodkin/pyzmq/blob/asyncmonitorwrap/zmq/utils/monitor.py

minrk: Yes, that looks right!

shimondoodkin:
I tried to do typing to the methods i tried the union thing with the return type and variations of awaitable and the return type, And it did not work.
Seems mypy / typeshed-paramikois don't work well yet with async functions. and typed collections validations are specific to python 3.10 which is too new.

JeffWScott:
I this PR going to get merged? I need this functionality.

minrk:
That's right - this is a convenience method, you definitely don't need it to handle monitor messages with async sockets:

```python
from zmq.utils.monitor import parse_monitor_msg
frames = await socket.recv_multipart()
msg = parse_monitor_msg(frames)
```

This function only saves you one line.

I'll merge it when I have a chance to iron out the last typing bits.

minrk:
Thanks!

Issue #1697, opened 2022-05-09
Python 3.11 Support


blink1073:
I tried to enable Python 3.11 beta for `jupyter_server`, but got this error when building [`pyzmq`](https://github.com/jupyter-server/jupyter_server/runs/6350119667?check_suite_focus=true):

```
             gcc -pthread -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -DHAVE_SYS_UN_H=1 -Ibundled/zeromq/include -Izmq/utils -I/opt/hostedtoolcache/Python/3.11.0-beta.1/x64/include/python3.11 -c zmq/backend/cython/_device.c -o build/temp.linux-x86_64-cpython-311/zmq/backend/cython/_device.o
            zmq/backend/cython/_device.c:217:12: fatal error: longintrepr.h: No such file or directory
              217 |   #include "longintrepr.h"
                  |            ^~~~~~~~~~~~~~~
            compilation terminated.
            error: command '/usr/bin/gcc' failed with exit code 1
            [end of output]
```

FWIW it looks like that file still [exists](https://github.com/python/cpython/blob/main/Include/cpython/longintrepr.h)

minrk:
I've been waiting for Cython 0.29.29 to fix the 3.11 builds, but I'm going to make a prerelease with #1696 since that's been taking a while

blink1073:
Ah, thanks!  Yeah it makes sense for us to target Python 3.11 only in our prerelease job.

blink1073:
In [progress](https://github.com/jupyter-server/jupyter_server/runs/6352001282?check_suite_focus=true)

blink1073:
Great!

minrk:
23.0 is up

PR #1707, opened 2022-06-01
add Python 3.11 wheels

minrk:
behind CIBW_PRERELEASE_PYTHONS flag for now (wait for 3.11 rc)

Move most cibuildwheel config to pyproject.toml to reduce logic only in CI workflow files (reduces duplication)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,12 @@ jobs:
     parameters:
       build:
         type: string
+      skip:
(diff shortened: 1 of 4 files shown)


minrk:
It works! need to wait for upload, because 3.11 wheels shouldn't be uploaded before beta 4 at the earliest, [expected in July](https://peps.python.org/pep-0664/).

closes #1697

PR #1709, opened 2022-06-01
setup: Fix build with mingw toolchain

Biswa96:
This fixes the error: Microsoft Visual C++ 14.0 or greater is required.


minrk:
Failures are due to: https://github.com/actions/runner/issues/1923

Windows builds passed, which are the only ones affected, so merging.

PR #1718, opened 2022-06-10
Safer cinit in Context

### LINKED ISSUE #1717: Cannot subclass Context in v23
One can no longer subclass `zmq.Context` in a simple way. This worked in 22.x:

```python
import zmq

class Context(zmq.Context):
    x = None
    def __init__(self, *args, x=7, **kwargs):
        self.x = x
        super().__init__(*args, **kwargs)

ctx = Context(x=6)
```

But raises an exception in 23.x:
```python
  File "/home/bilbo/Documents/Scratch/980.py", line 9, in <module>
    ctx = Context(x=6)
  File "zmq/backend/cython/context.pyx", line 36, in zmq.backend.cython.context.Context.__cinit__
TypeError: __cinit__() got an unexpected keyword argument 'x'
```
This is due to [this change](https://github.com/zeromq/pyzmq/commit/2dc67d544c8fc853c9acca3cef7f89803d92298a#diff-00887357f259afc1499eb5b42a2f1e5d5f7c6102b68e6b1478c4ebc7df1dd0c7), which modified the cython class for Context's `__cinit__` method to no longer ignore additional keyword arguments passed to a subclass.

```diff
-     def __cinit__(self, int io_threads=1, size_t shadow=0, **kwargs):
+     def __cinit__(self, int io_threads=1, size_t shadow=0):
```

One can still make a subclass, but it now requires overriding `__new__` to ensure the extra arguments aren't passed to the parent's `__new__` (which is where `__cinit__` gets called - I admit I'm a little hazy on the details). This seems to work:

```python
class Context(zmq.Context):
    x = None
    def __new__(cls, *args, x=7, **kwargs):
        return super().__new__(cls, *args, **kwargs)

    def __init__(self, *args, x=7, **kwargs):
        self.x = x
        super().__init__(*args, **kwargs)
```

I can of course modify my subclasses as above,. but thought you might want to know about this in case you consider it a regression.

minrk:
Matches what we do in Socket for the same reason - most logic is actually in `__init__`, not `__cinit__`. This leaves subclasses to be more flexible with their signatures.

closes #1717

### DIFF
--- a/zmq/backend/cython/context.pyx
+++ b/zmq/backend/cython/context.pyx
@@ -29,12 +29,12 @@ cdef class Context:
         The number of IO threads.
     """
 
-    # no-op for the signature
-    def __init__(self, io_threads=1, shadow=0):
-        pass
-
-    def __cinit__(self, int io_threads=1, size_t shadow=0):
+    def __cinit__(self, *args, **kwargs):
         self.handle = NULL
+        self._pid = 0
+        self._shadow = False
+
+    def __init__(self, int io_threads=1, size_t shadow=0):
         if shadow:
             self.handle = <void *>shadow
             self._shadow = True
--- a/zmq/tests/test_ext.py
+++ b/zmq/tests/test_ext.py
@@ -1,0 +1,34 @@
+"""tests for extending pyzmq"""
+
+import zmq
+
+
+class CustomSocket(zmq.Socket):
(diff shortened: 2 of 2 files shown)


chrisjbillington:
Thanks for the quick fix! I can confirm this (unsurprisingly) resolves the issue for me.

PR #1760, opened 2022-09-10
Support async callback in authenticator

### LINKED ISSUE #1729: zmq.auth.asyncio.AsyncioAuthenticator - enhancement proposal
# Whats up?
The ability to avoid blocking the main event loop on I/O ops using python's asyncio lib is hugely important to our project. So is being able to verify if a connection attempt from a peer is one we are happy to allow. 

Currently we are using pyzmq's AsyncioAuthenticator to handle authenticating peers before allowing them to send/recv messages. However we've hit a problem. Turns out that [how zap messages are handled](https://github.com/zeromq/pyzmq/blob/58105c35394c3c13ec762965771daf2384e92df0/zmq/auth/asyncio.py#L40) doesn't allow for using await able callbacks, and this means we have to use a threaded version of the authenticator instead.

## In detail
The way we have currently setup our means of authentication is that we configure curve security using a [CredentialsProvider class](https://pyzmq.readthedocs.io/en/latest/api/zmq.auth.html#zmq.auth.Authenticator.configure_curve_callback) with a custom implementation of the callback method that queries a hierarchy of resources to check whether we `like` the public key of the peer that is attempting to connect. 

One of those resources is a web service, so we initially thought of using the AsyncioAuthenticator so we can be consistent with other queries to our same web service which also will be using asyncio. We had planned to design a service that would do this work and that the callback method would use that service. 

However we're now facing a split here since it became clear that the handling of callbacks in AsyncioAuthenticator doesn't actually treat the CredentialsProvider.callback method as an async method. It calls it as a normal method. So we cannot use our asyncio based web service to make queries about authentication. 

This is awkward to say the least as we either need to redesign the whole web query service, or have a different, special service for just the auth part. 

## What could be done
In my opinion, I see no reason why the callback method of the CredentialsProvider wouldn't be expected to do things like query databases, either locally or remotely. These seem like the very things asyncio was designed to cater for without blocking the main thread. 

I'd like to propose a PR that would enable two things;
1. Keep the existing API consistent with how it is now, so as not to break compatibility with anyone who's implemented using the AsyncioAuthenticator with a non-awatiable callback.
2. Insert a check to inspect if the callback is await able or not, if it is await able, then cause the AsyncioAuthenticator to await the result of the callback method of the CredentialsProvider so that it won't block the main thread/event loop. 

Having taken a look at the code, though it would take a smidge of refactoring, there don't appear to be any serious issues in implementing this as far as I can tell. 

## What I ask
A simple yes, go for it, or, no, and this is why not answer to my question of can we do a PR to fix this issue :)

pembem22:
This PR adds support for async callbacks in authenticator by making the base `handle_zap_message` method async, and the thread implementation run the asyncio authenticator in a background thread, as suggested in the issue.

Closes #1729


pembem22:
Hi @minrk, the background thread in threaded authenticator uses asyncio poller to receive commands from main thread, similar to how asyncio authenticator polls the zap socket. Right now I'm stuck with some kind of a deadlock issue. The main thread hangs when trying to send a message through the pipe socket and the background one hangs when awaiting for poller. Could you help me solve this?

minrk:
asyncio cannot be used safely across threads. That means you can't use an asyncio poller or socket in more than one thread. You can use _different_ async objects across threads, communicating with zmq. My guess is that objects are being attached to event loops multiple times.

Ideally, async objects will not be created outside a running event loop, and should never be passed from one loop to another.

pembem22:
I create and set another asyncio event loop inside the thread [here](https://github.com/zeromq/pyzmq/pull/1760/files#diff-2de56d65ed3519199b3e86bdd9da3eb3bd41be9710c4b5c76c2f758385cc0edbR54) and only after that I create new socket, poller, and authenticator. I don't see any asyncio objects being shared between threads or loops, so not sure what could be the issue.

zmq/auth/thread.py
```diff
@@ -6,101 +6,92 @@
...
     def __init__(
         self,
-        context: "zmq.Context",
         endpoint: str,
         encoding: str = 'utf-8',
         log: Any = None,
-        authenticator: Optional[Authenticator] = None,
     ) -> None:
         super().__init__()
-        self.context = context or zmq.Context.instance()
```
minrk, line 36: context ~always has to be passed here, because the transport is `inproc://`, the (underlying libzmq) context must be the same. I think that's why you're getting hangs (send on a socket with inproc will hang until the peer arrives).

I need to think of a better way to do this, but you can make an async reference to an existing context with:

```python
actx = zmq.asyncio.Context.shadow(ctx.underlying)
async_socket = actx.socket(zmq.PULL)
```

Next, you'll really want to do everything async within a coroutine, so `run` should look like:

```python
    def run(self) -> None:
        """Start the Authentication Agent thread task"""
        loop = asyncio.new_event_loop()
        try:
            loop.run_until_complete(self._run())
        finally:
            loop.close()
            # cleanup sockets

    async def _run(self) -> None:
        context = zmq.asyncio.Context.shadow(self.context.underlying)
        ...

```

pembem22:
Thanks, that fixed the hanging issue! The only thing left now is to await the callback in tornado ioloop authenticator. That requires changing the `_run_callback` method of `zmq.eventloop.zmqstream.ZMQStream` [here](https://github.com/zeromq/pyzmq/blob/main/zmq/eventloop/zmqstream.py#L557) to await the result, but I can't figure that out.

minrk:
#1785 fixed support for async callbacks in zmqstream, if you want to update this PR and see if that's enough.

pembem22:
#1785 fixes all tests for me locally and I also added tests with async callback. But some tests are failing here, not sure how to debug them.

minrk:
My guess is it's because `deny` is a not-awaited async method, and something's changed in how long it takes to be handled. It's always had a race, but apparently something in the changes here make it more likely that the auth thread loses the race.

Adding a reply to the pipe messages should make it easy to be sure deny doesn't return before it's had an effect. I'll have a look.

minrk:
I think the tests failing are because of wonky things in the test runners, not the changes here. I'm in the middle of rewriting those tests now, so I'll merge this and finish that up a bit later.

Thanks!

PR #1769, opened 2022-09-19
Use weakref.finalize to fix issue with warnings.warn being None during shutdown.

DrTodd13:
I ran into an issue where the warnings.warn was None so the call will generate an exception (which is then ignored since it is in a __del__).  I think this is based on some other package we were using that made that change during shutdown.  This change to use finalize instead lets the finalizer run earlier before the change to warn.

### DIFF
--- a/zmq/sugar/socket.py
+++ b/zmq/sugar/socket.py
@@ -8,6 +8,7 @@ import errno
 import pickle
 import random
 import sys
+import weakref
 from typing import (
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

Since this significantly changed the life cycle of some things, I figured handling the global `warn` being None was a safer, smaller change. A few similar improvements to finalization added in #1770

PR #1822, opened 2023-01-02
start publishing manylinux wheels for both glibc 2.17 and 2.28, starting with cpython 3.10

### LINKED ISSUE #1821: Wheel for Python 3.11 and x86_64, manylinux_2_17 becomes manylinux_2_28 - an issue?
During `pip install pyzmq` in a Centos 7 based docker image with Python 3.10 and Python 3.11 setup respectively, pip switches from downloading a wheel to downloading a source distribution. The reason is that for Python 3.11 of the x86_64 arch specifically, the published pyzmq wheel switched from manylinux_2_17 to 2_28.

Is this a constraint stemming from pyzmq, py311, or pyzmq's release automation as configured using cibuildwheel (recently https://github.com/zeromq/pyzmq/pull/1707), or something else?

I tried to figure those details out myself but got stuck, so watch out - this issue can be off topic for pyzmq as a project.

### Centos 7 with Python 3.10, pip install pyzmq

```shell
Collecting pyzmq>=17
  Downloading pyzmq-24.0.1-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (1.1 MB)
```
image

### Centos 7 with Python 3.11, pip install pyzmq

```shell
Collecting pyzmq>=17
  Downloading pyzmq-24.0.1.tar.gz (1.2 MB)
```
image

## Related

- https://pypi.org/project/pyzmq/25.0.0b1/#files
- https://peps.python.org/pep-0599/
- https://peps.python.org/pep-0600/

minrk:
rather than only having one choice specified for each Python build in pyproject.toml (2.17 for 3.10, 2.28 for 3.11), build everything starting at 3.10 with manylinux2014. Then _also_ build everything starting with 3.10 with manylinux_2_28.

cibuildwheel doesn't support multiple values for the image inputs, so 2014 is the default and 2_28 is added via a gha matrix entry.

These future-glibc builds are published only for x86_64.

The change only adds builds (all x86_64):

- 2014 for cp311
- 2_28 for cp310, pp39

The rest should be unchanged.

closes #1821

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -78,6 +78,7 @@ jobs:
       CIBW_SKIP: "${{ matrix.cibw.skip || '' }}"
       CIBW_ARCHS_LINUX: "${{ matrix.cibw.arch || 'auto' }}"
       CIBW_ARCHS_MACOS: "${{ matrix.cibw.arch || 'auto' }}"
+      CIBW_MANYLINUX_X86_64_IMAGE: "${{ matrix.cibw.manylinux_x86_64_image || '' }}"
 
     strategy:
       fail-fast: false
@@ -110,6 +111,15 @@ jobs:
             cibw:
               arch: i686
               build: "*manylinux*"
+
+          # additional manylinux variants, not specified in pyproject.toml:
+          # build with newer 2_28 for cpython >= 3.10, pypy 3.9
+          - name: manylinux-x86_64-2_28
+            os: ubuntu-20.04
+            cibw:
+              arch: x86_64
+              build: "cp31*-manylinux* pp39-manylinux*"
+              manylinux_x86_64_image: manylinux_2_28
 
           - os: ubuntu-20.04
             name: musllinux
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,6 +64,8 @@ test-command = "pytest -vsx {package}/tools/test_wheel.py"
 [tool.cibuildwheel.linux]
 before-all = "bash tools/install_libzmq.sh"
 
+manylinux-x86_64-image = "manylinux2014"
+manylinux-i686-image = "manylinux2014"
 manylinux-aarch64-image = "manylinux2014"
 musllinux-aarch64-image = "musllinux_1_1"
 musllinux-i686-image = "musllinux_1_1"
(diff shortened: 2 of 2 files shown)


.github/workflows/wheels.yml
```diff
@@ -78,6 +78,7 @@ jobs:
       CIBW_SKIP: "${{ matrix.cibw.skip || '' }}"
       CIBW_ARCHS_LINUX: "${{ matrix.cibw.arch || 'auto' }}"
       CIBW_ARCHS_MACOS: "${{ matrix.cibw.arch || 'auto' }}"
+      CIBW_MANYLINUX_IMAGE: "${{ matrix.cibw.image || '' }}"
```
consideRatio, line 81: I'm not sure you can set this environment variable, I understand that you could set `CIBW_MANYLINUX_X86_64_IMAGE` for example though from https://cibuildwheel.readthedocs.io/en/stable/options/.

Have you confirmed this works?

.github/workflows/wheels.yml
```diff
@@ -111,6 +112,15 @@ jobs:
...
               build: "*manylinux*"
 
+          # additional manylinux variants, not specified in pyproject.toml:
+          # build with newer 2_28 for cpython >= 3.10, pypy 3.9
+          - name: manylinux-x86_64-2_28
+            os: ubuntu-20.04
+            cibw:
+              arch: x86_64
+              build: "cp31*-manylinux* pp39-manylinux*"
+              image: manylinux_2_28
```
consideRatio, line 122: I think the `image` key is a too brief description since its configuring specifically a manylinux associated environment variable. I suggest re-using exactly the name of the env it will configure to reduce complexity of understanding whats going on.

consideRatio (commented):
Ah, we worked on this at the same time. Which is fine! I wanted to learn a bit about cibuildwheel.

Your strategy seems better, I'll close my PR!

.github/workflows/wheels.yml
minrk: Good suggestion! Done.

.github/workflows/wheels.yml
minrk: Good catch! It definitely does not.  I _think_ it used to a long time ago, but I've definitely made this exact mistake multiple times before.

consideRatio (approved):
Looks great! Thanks @minrk!!!

minrk:
Double checked the expected wheels are produced (I don't have an easy test for this, but it changes so rarely I'm okay with that). Thanks for reviewing!

PR #1832, opened 2023-02-02
Add top level license files

dcsaba89:


minrk:
these files are already in the top-level of the repo as COPYING.BSD, COPYING.LESSER

dcsaba89:
Hi @minrk, that file could not be recognized as a license file and causing issues for us, that is why I commited a new license file based on the following guide:
https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-license-to-a-repository

Please take a look at other repositories, the license file is in the root of the repository and called LICENSE or maybe LICENSE.txt

https://github.com/pandas-dev/pandas/blob/main/LICENSE
https://github.com/numpy/numpy/blob/main/LICENSE.txt
https://github.com/ipython/ipython/blob/main/LICENSE
https://github.com/jupyter/nbclassic/blob/main/LICENSE
https://github.com/jupyter-widgets/ipywidgets/blob/main/LICENSE

And you can see a badge on the top of the pages that shows the license itself and the associated permissions, limitations and conditions as well.

However you cannot see the same badge for your top level license files
https://github.com/zeromq/pyzmq/blob/main/COPYING.BSD
https://github.com/zeromq/pyzmq/blob/main/COPYING.LESSER

This change would help significantly to update pyzmq more frequently for our python projects.

PR #1835, opened 2023-02-17
handle custom Message types in zmq.log

### LINKED ISSUE #1834: zmq logger does not support custom Message types
### What pyzmq version?

25.0.0

### What libzmq version?

4.3.4

Python 3.10.8 | packaged by conda-forge | (main, Nov 22 2022, 08:26:04) [GCC 10.4.0] on linux

### OS

Ubuntu 22.04

### What happened?

Apparently, the [`emit`](https://github.com/zeromq/pyzmq/blob/9b767422dd8b32c29496d5fe1858b4e325fb42d2/zmq/log/handlers.py#L132-L157) function in pyzmq's PubHandler doesn't take into account the possibility of typecasting when applying the `split` function for a LogRecord's msg attribute in python. This causes a lot of issues when the log handler is used in conjunction with the logger already in existence. A possible resolution might be to use [`getMessage`](https://docs.python.org/3/library/logging.html#logging.LogRecord.getMessage) instead which does the type casting.
 
https://github.com/zeromq/pyzmq/blob/9b767422dd8b32c29496d5fe1858b4e325fb42d2/zmq/log/handlers.py#L132-L157

### Code to reproduce bug

```python
import random
import time

import zmq.log.handlers
import logging


class DummyClass:
    def __init__(self, prop1: int) -> None:
        self.prop1 = prop1

    def __str__(self):
        return f"DummyClass(prop1={self.prop1})"


class DummyLogServer:
    def __init__(self, port=10000):
        self.port = port
        self.context = zmq.Context()
        self.socket = self.context.socket(zmq.PUB)
        self.socket.bind("ws://*:%s" % self.port)

        self.logger = logging.getLogger("DummyLogServer")
        self.logger.setLevel(logging.DEBUG)
        formatter = logging.Formatter(
            "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
        )

        self.pub_handler = zmq.log.handlers.PUBHandler(self.socket)
        self.pub_handler.setFormatter(formatter)
        self.console_handler = logging.StreamHandler()

        self.pub_handler.root_topic = "mock_logs"
        self.logger.addHandler(self.pub_handler)
        self.logger.addHandler(self.console_handler)

    def run(self):
        while True:
            callable = random.choice(
                [
                    self.logger.debug,
                    self.logger.info,
                    self.logger.warning,
                    self.logger.error,
                    self.logger.critical,
                ]
            )
            callable(
              DummyClass(random.randint(0, 2000))
            )
            time.sleep(random.random())


if __name__ == "__main__":
    server = DummyLogServer()
    server.run()
```

### Traceback, if applicable

```shell
file "/path/to/python_lib/python3.10/site-packages/zmq/log/handlers.py", line 136, in emit
    topic, record.msg = record.msg.split(TOPIC_DELIM, 1)
AttributeError: 'DummyClass' object has no attribute 'split'
```

### More info

_No response_

minrk:
closes #1834

### DIFF
--- a/zmq/log/handlers.py
+++ b/zmq/log/handlers.py
@@ -21,6 +21,7 @@ Code adapted from StarCluster:
 """
 
 import logging
+from copy import copy
 
 # Copyright (C) PyZMQ Developers
 # Distributed under the terms of the Modified BSD License.
@@ -132,10 +133,16 @@ class PUBHandler(logging.Handler):
     def emit(self, record):
         """Emit a log message on my socket."""
 
+        # LogRecord.getMessage explicitly allows msg to be anything _castable_ to a str
         try:
-            topic, record.msg = record.msg.split(TOPIC_DELIM, 1)
+            topic, msg = str(record.msg).split(TOPIC_DELIM, 1)
         except ValueError:
             topic = ""
+        else:
+            # copy to avoid mutating LogRecord in-place
+            record = copy(record)
+            record.msg = msg
+
         try:
             bmsg = self.format(record).encode("utf8")
         except Exception:
@@ -152,7 +159,7 @@ class PUBHandler(logging.Handler):
         if topic:
             topic_list.append(topic)
 
-        btopic = '.'.join(topic_list).encode("utf8")
+        btopic = '.'.join(topic_list).encode("utf8", "replace")
 
         self.socket.send_multipart([btopic, bmsg])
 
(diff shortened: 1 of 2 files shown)


minrk:
Turns out getMessage wasn't quite right, since that's called in _format_, and I don't think we can call format before stripping the topic. So instead, call str explicitly, just like in getMessage, rather than calling getMessage itself.

davetapley:
@minrk thanks for quick fix, can we get a new release? 🙏🏻

PR #1837, opened 2023-02-17
Clean up license files

dcsaba89:
replace COPYING.BSD and COPYING.LESSER with LICENSE.GNU and LICENSE.LESSER respectively

@blink1073, same change as the below PRs for jupyter-client for which now the package's license is perfectly recognized by github as well

https://github.com/jupyter/jupyter_client/pull/934

please approve


minrk:
Thanks!

COPYING.LESSER
```diff
@@ -1,181 +0,0 @@
...
-General Public License, you may choose any version of the GNU Lesser
-General Public License ever published by the Free Software Foundation.
-
-  If the Library as you received it specifies that a proxy can decide
-whether future versions of the GNU Lesser General Public License shall
-apply, that proxy's public statement of acceptance of any version is
-permanent authorization for you to choose that version for the
-Library.
-
---------------------------------------------------------------------------------
```
minrk, line 167: For consistency, we need to make sure this static-linking exception from the original authors is preserved.

dcsaba89: Thanks @minrk, done.

PR #1838, opened 2023-02-22
update documentation page

12rambau:
I'm a maintainer of the pydat-sphinx-theme (that you are using to build your documentation. I was passing by and I think you have some legacy code polluting the final ouptut This PR will:

- remove the templated headerbar (I assume it was for the RDT theme) that is not needded anymore
- use fontawesome 6 icons
- set the logo in the navbar
- remove unused extention

I think it would make sense to remove all the commented content of the conf.py file but it's as you wich


minrk:
Thanks!

PR #1840, opened 2023-02-23
allow `socket` or `url` as keyword args to PUBHandler

### LINKED ISSUE #1839: DOC: Can't use PUBHandler with dictConfig
### What pyzmq version?

24.0.1

### What libzmq version?

4.3.4

3.10.9 via pyenv

### OS

ubuntu 20.04

### What happened?

I use [`dictConfig`](https://docs.python.org/3/library/logging.config.html#logging.config.dictConfig) to configure my other loggers, but I needed a workaround (see below) to make it work with `PUBHandler`.

This happens because the `PUBHandler` constructor takes `socket` as an arg, but with `dictConfig`:

> All other keys are passed through as keyword arguments to the handler’s constructor.

### Code to reproduce bug

```python
from logging.config import dictConfig

from zmq import PUB, Context

socket = Context.instance().socket(PUB)
socket.connect('tcp://whatever')

config = {
    'version': 1,
    'handlers': {
        'zmq': {
            'class': 'zmq.log.handlers.PUBHandler'
            'socket': socket
        }
    }
}

dictConfig(config)
```

### Traceback, if applicable

```shell
TypeError: PUBHandler.__init__() got an unexpected keyword argument 'socket'
The above exception was the direct cause of the following exception:

...

ValueError: Unable to configure handler 'zmq'
```

### More info

Workaround is to create:

```py
class DictConfigPUBHandler(PUBHandler):
    def __init__(self, **kwargs):
        socket = kwargs['socket']
        del kwargs['socket']
        super().__init__(socket, **kwargs)
```

Then use:

```py
        'zmq': {
            'class': 'path.to.DictConfigPUBHandler'
            'socket': Context.instance().socket(PUB)
```

minrk:
closes #1839

### DIFF
--- a/zmq/log/handlers.py
+++ b/zmq/log/handlers.py
@@ -59,9 +59,12 @@ class PUBHandler(logging.Handler):
 
     def __init__(
         self,
-        interface_or_socket: Union[str, zmq.Socket],
+        interface_or_socket: Optional[Union[str, zmq.Socket]] = None,
         context: Optional[zmq.Context] = None,
         root_topic: str = '',
+        *,
+        socket: Optional[zmq.Socket] = None,
+        url: Optional[str] = None,
     ) -> None:
         logging.Handler.__init__(self)
         self.root_topic = root_topic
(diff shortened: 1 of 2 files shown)


minrk:
This complicates things, so I'm not sure if it's actually an improvement over specifying the current argument name `interface_or_socket`, [as documented](https://pyzmq.readthedocs.io/en/latest/api/zmq.log.handlers.html?highlight=interface_or_socket#zmq.log.handlers.PUBHandler)

minrk:
Closing in favor of a dictConfig example in the docs instead.

PR #1848, opened 2023-03-16
handle closed shadow sockets in ZMQStream

minrk:
I believe this handles the case in https://github.com/spyder-ide/spyder/issues/20381

cc @ccordoba12


ccordoba12:
Looks good to me, thanks @minrk!

PR #1879, opened 2023-06-24
Advertise support up to Python 3.12

lod:
The CI is building up to 3.12, so it is tested

Advertise it for users and to trigger pypi to package wheels


minrk:
Thanks! I'm curious - how do you use PyPI classifiers that this came up? They don't seem very useful to me (particularly the version support ones, where their values can often change without needing a new release), and usually end up choosing `Python 3`.

lod:
I was installing into a python:3.11 container and it wanted to compile the package, then it got sad because it didn't have a compiler

pypi uses the classifiers to determine which wheels to build and distribute

Which is super convenient for container usage

minrk:
I'm not sure I understand. PyPI doesn't build pyzmq wheels, I do, and there are wheels (which pip will get by default) up to Python 3.12.

This works today, while there are still no classifiers on PyPI:

```dockerfile
FROM python:3.12.0b3
RUN python3 -mpip install --pre pyzmq # gets wheel
```

Note: you do need to use `--pre`, as there are only prerelease wheels with 3.12 support right now, but stable versions of Python get wheels now, too.

PR #1894, opened 2023-08-10
libzmq 4.3.5 removed proxy_steerable for license reasons

minrk:


bluca:
@minrk FYI proxy_steerable has been re-added, before the latest release

PR #1903, opened 2023-10-11
bundle libzmq 4.3.5

minrk:
libzmq no longer supports binaries for Windows, and has removed support for tweetnacl.

This may mean that Windows wheels will lose CURVE, or we have to get binaries elsehwere, or run our own builds of libzmq to bundle, which I suspect would be _highly_ likely to get the wrong DLL dependency bundling.

Maybe related to #1902 since windows arm builds ought to work, I think?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,45 +33,45 @@ jobs:
       fail-fast: true
       matrix:
         include:
-          - os: macos-13
-            python: 3.7
-
-          - os: macos-13
-            python: 3.11
-            zmq: bundled
-
-          - os: macos-13
-            python: pypy-3.9
-            zmq: bundled
-
-          - os: ubuntu-20.04
-            python: 3.7
-            zmq: bundled
-            tornado: none
-
-          - os: ubuntu-22.04
-            python: pypy-3.9
-            zmq: bundled
-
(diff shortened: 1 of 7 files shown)


minrk:
Increasingly, I think the way we do bundling should change due to this, but what we have now is so brittle and works in so many untestable cases that I dread it.

I think:

1. it should download and install libsodium and libzmq, not bundle sources
2. those installers should use their _own_ installation scripts (./configure or cmake), not setup.py
3. (maybe) take the opportunity to move to a modern build tool, like [scikit-build-core](https://scikit-build-core.readthedocs.io) or [meson](https://meson-python.readthedocs.io/en/latest/).

Pros:

- This would eliminate much of the complexity of setup.py (trading it for as-yet-unidentified complexity of new tools)
- Running their own builds, would eliminate maintenance of the `include_{platform}` files
- Download & install means no more bundling sources

Cons:

- New build tooling will _inevitably_ break some installs, and cause maintenance trouble for some time, especially since I have hardly any experience with scikit-build-core, and none with meson
- Downloading external dependencies at install time means pyzmq can't install from a single tarball without Internet _if_ libzmq is not found (decreasingly relevant, but not _irrelevant_)
- I don't really have time for this

PR #1905, opened 2023-10-12
Changed openssh password prompt regex + Added max number of retries ; fix #1904

### LINKED ISSUE #1904: BUG: `try_passwordless_ssh` hangs while connecting to server
### What pyzmq version?

23.2.0

### What libzmq version?

4.3.4

Python 3.9.13, from Anaconda

### OS

Ubuntu 22.04

### What happened?

The function `try_passwordless_ssh` should return `False` but instead, it hangs indefinitely.

### Code to reproduce bug

```python
import zmq.ssh

SERVER_IP = "123.123.123.123" 
KEY = "/path/to/my/key"
zmq.ssh.try_passwordless_ssh(SERVER_IP, KEY)
# Hangs there indefinitely
```

### Traceback, if applicable

_No response_

### More info

Came here from [Helium](https://github.com/sschuhmann/Helium) sublime-text plugin which uses pyzmq under the hood. However, I was able to replicate it with Anaconda.

Related to #676, albeit not fixed by the fix proposed for that issue.

There seems to be 2 problems with the code for the function:

  1. `try_passwordless_ssh` has a `while True:` loop which always has potential to hang if the `ssh` command doesn't behave like `pyzmq` expects ; some max retry number should be added in there.
  2. My `ssh` program (shipped with Ubuntu 22.04) doesn't behave like `pyzmq` expects. It has the passphrase prompt below. By contrast, the `_password_pat` used in `try_passwordless_ssh` wants "passphrase" to be immediately followed by ":"

```
Enter passphrase for key '/path/to/my/key/`: 
```

I can beautify the hacky fix I have and submit it as a pull request.

KenyC:
My `ssh`  doesn't have the password prompt `pyzmq` expects, which triggers infinite looping (as described in #1904).

```
> ssh -f 123.123.123.123 -i /path/to/my/key/
Enter passphrase for key '/path/to/my/key/': 
> ssh -V
OpenSSH_8.9p1 Ubuntu-3ubuntu0.4, OpenSSL 3.0.2 15 Mar 2022
```

I changed the pattern from `pass(word|phrase):` to `pass(word|phrase)` to account for my `ssh`.

I also added a max number of  retry to avoid infinite looping, in case future updates also change the phrasing of password prompts.

closes #1904

### DIFF
--- a/zmq/ssh/tunnel.py
+++ b/zmq/ssh/tunnel.py
@@ -39,6 +39,10 @@ except ImportError:
     pexpect = None
 
 
+class MaxRetryExceeded(Exception):
+    pass
+
+
 def select_random_ports(n):
     """Select and return n random ports that are available."""
     ports = []
@@ -56,7 +60,7 @@ def select_random_ports(n):
 # -----------------------------------------------------------------------------
 # Check for passwordless login
 # -----------------------------------------------------------------------------
-_password_pat = re.compile(rb'pass(word|phrase):', re.IGNORECASE)
+_password_pat = re.compile(rb'pass(word|phrase)', re.IGNORECASE)
 
 
 def try_passwordless_ssh(server, keyfile, paramiko=None):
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

PR #1921, opened 2024-01-02
Replace setsockopt() by setsockopt_string()

radeksm:
Fix below error:
  TypeError: unicode not allowed, use setsockopt_string

And remove remaining references to socket.setsockopt() and replace by socket.setsockopt_string().

### DIFF
--- a/docs/source/notes/unicode.md
+++ b/docs/source/notes/unicode.md
@@ -173,14 +173,6 @@ Overview of the relevant methods:
 ```
 
 ```{eval-rst}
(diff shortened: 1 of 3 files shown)


minrk:
Thanks for the PR! This is not a correct change, the setsockopt interface is bytes, and most often integers.

If I'm reading correctly, the TypeError is correct, in that if you give unicode you've made an error, and you should be giving it bytes.

radeksm:
@minrk  yeahhh you are correct. Thanks and sorry !

PR #1922, opened 2024-01-05
WIP meson

minrk:
try switching build system to meson, which lets us bundle 'true' builds of libzmq/libsodium.

builds on #1903, because that's why we need it, since libsodium becomes a dependency of bundled libzmq due to the removal of libzmq's tweetnacl.

So far, the simple  Cython + CPython on linux/mac is working

TODO:

- [x] add meson.build
- [x] preserve environment variables for picking bundled libzmq
- [x] statically link libzmq and libsodium when libzmq is not found
- [x] build Cython backend
- [x] build CFFI backend on PyPy
- [ ] build static libs on Windows with msbuild, cmake
- [ ] enable adequate configuration for the many cases we supported previously. Meson's philosophical objection to env vars makes this much harder than it needs to be.

[meson discussion](https://github.com/mesonbuild/meson-python/discussions/556)

PR #1927, opened 2024-02-14
build with cmake (scikit-build-core)

minrk:
not sure if meson is a good fit, so going to try scikit-build-core, which seems a lot simpler and less aggressive

seems to have been a lot easier than meson in #1922. I'm not great at cmake, but since libzmq uses it, integration was easier.

closes #1922
closes #1903


dimpase:
this seems to be breaking SageMath build, as scikit does something something ... https://github.com/cschwan/sage-on-gentoo/issues/783

PR #1944, opened 2024-02-20
update draft install doc

minrk:
remove discussion of prerelease arg, not required since pyzmq 17 was released  in 2018.

cc @pabera

### DIFF
--- a/docs/source/howto/draft.md
+++ b/docs/source/howto/draft.md
@@ -31,10 +31,10 @@ And then build pyzmq with draft support:
 ```bash
 export ZMQ_PREFIX=${PREFIX}
 export ZMQ_DRAFT_API=1
-pip install -v --no-binary pyzmq pyzmq
+pip install -v pyzmq --no-binary pyzmq
 ```
 
-By specifying `--no-binary pyzmq`, pip knows to not install wheels, and will compile pyzmq from source.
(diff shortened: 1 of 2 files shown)


docs/source/howto/draft.md
```diff
@@ -26,19 +26,19 @@ cd zeromq-${ZMQ_VERSION}
...
 ```
 
-And to install pyzmq with draft support:
+And then build pyzmq with draft support:
 
 ```bash
 export ZMQ_PREFIX=${PREFIX}
 export ZMQ_DRAFT_API=1
-pip install -v --no-binary pyzmq --pre pyzmq
+pip install -v --no-binary pyzmq pyzmq
```
pabera, line 34: I am not a `pip` expert, but shouldn't this line be either just

`pip install -v --no-binary pyzmq`

or 

`pip install -v --no-binary :pyzmq: pyzmq`

or

`pip install -v --no-binary :all: pyzmq`

minrk: Thanks for checking! The line is correct. `--no-binary` takes an argument for which package(s) to install from source, and then remaining package names are what to install. `:all:` is a special string that's not a valid package name, meaning install no binaries at all, which you don't actually want because that prevents e.g. installing wheels of build dependencies like `cython`. Since pyzmq is the only package you want to install from source, you want `--no-binary pyzmq`.

pabera: Make sense. Thank you for the explanation.

PR #1972, opened 2024-04-07
tests: using await for on_send(), on_recv() with ZMQStream

spchamp:
Update: This patch may be an artifact of an incomplete test installation. `pytest_asyncio` was not installed under the initial testing environment.  This may have resulted in the test failure I'd seen.

Maybe it's possible to await some things related to a tornado IO loop, this patch might not in fact be the right approach though


spchamp:
Testing this again with the unpatched sources now that `pytest_asyncio` is installed, the tests pass. I'll delete this branch shortly.

Issue #1978, opened 2024-04-19
BUG: pyzmq 26.0.1 failed to install


crazygao:
### This is a pyzmq bug

- [X] This is a pyzmq-specific bug, not an issue of zmq socket behavior. Don't worry if you're not sure! We'll figure it out together.

### What pyzmq version?

26.0.1

### What libzmq version?

Nan

### Python version (and how it was installed)

every version

### OS

Every system

### What happened?

![image](https://github.com/zeromq/pyzmq/assets/2208599/07acd84b-6a14-4840-b7fe-861ebfc02cdf)

jorenham:
I'm facing the same issue with python-poetry:

```
❯ poetry update
Updating dependencies
Resolving dependencies... (5.0s)

Package operations: 0 installs, 2 updates, 0 removals

  - Updating pyzmq (26.0.0 -> 26.0.1): Failed

  RuntimeError

  Unable to find installation candidates for pyzmq (26.0.1)

  at ~/.local/share/pypoetry/venv/lib/python3.11/site-packages/poetry/installation/chooser.py:74 in choose_for
       70│ 
       71│             links.append(link)
       72│ 
       73│         if not links:
    →  74│             raise RuntimeError(f"Unable to find installation candidates for {package}")
       75│ 
       76│         # Get the best link
       77│         chosen = max(links, key=lambda link: self._sort_key(package, link))
       78│ 

Cannot install pyzmq.
```

L1Aigner:
Facing the same problem with my Poetry setup. Unfortunately, before this error, I didn't even know I use this package somewhere. 😄 
Tbh I don't want to restrict the version now explicitly within my `pyproject.toml` but it is also braking my GitHub actions. I'm looking forward to a hotfix somewhere 😄

Matesanz:
## How to reproduce:

:point_right:  **Command:**

```bash
pip install pyzmq==26.0.1
```

:x:  **Error**

```
[...]

PR #1979, opened 2024-04-19
Update wheels.yml to prevent release if failed builds. 

markclare1992:
Add check for all wheels building successfully before uploading to PyPI. Not sure if this step should happen immediately after wheel building step or where I have put it?

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -234,5 +234,11 @@ jobs:
         with:
           path: dist
           merge-multiple: true
+      - name: Verify all wheel builds are successful
+        run: |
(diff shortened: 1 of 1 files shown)


markclare1992:
Linked to https://github.com/zeromq/pyzmq/issues/1978

minrk:
Thanks for the PR! This is already accomplished by the `needs` field, I believe (the affected step didn't run until just now when all the jobs finished). The problem is linux ARM wheels are published from CircleCI.

If we added a step that pulled the wheels in from Circle so that they all uploaded at once, it would fix this.

minrk:
#1982 causes all wheels to be uploaded together

PR #1989, opened 2024-05-16
chore: minor correction in required CMake

henryiii:
The minimum require version of CMake for scikit-build-core is 3.15, so setting it below 3.15 isn't helpful.

It could be removed altogether, as the default it to require what scikit-build-core requires, but leaving it here might make it easier to update on your end if you decide to raise it someday (3.26 is really nice!)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,8 +8,6 @@ requires = [
 ]
 build-backend = "scikit_build_core.build"
 
-# build-backend = "setuptools.build_meta"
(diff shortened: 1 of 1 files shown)


minrk:
Thanks! I'd like to require a more recent version! But I encountered some failures when cmake isn't available as a wheel, so wanted to be conservative and accept most versions that are out there. My needs are pretty basic beyond FetchContent.

PR #2033, opened 2024-09-12
try cython stable

minrk:
3.1 alpha is failing due to https://github.com/cython/cython/issues/6387


minrk:
nope, free-threading doesn't build with 3.0, maybe they just meant 'regular' 3.13.

PR #2045, opened 2024-10-29
Support AnyIO

### LINKED ISSUE #1827: FEAT: support AnyIO
### What pyzmq version?

25.0.0

### What libzmq version?

4.3.4

python 3.11.0, via conda-forge

### OS

ubuntu 22.04

### What happened?

I would like to use pyzmq with an async context inside an [AnyIO](https://anyio.readthedocs.io) event loop, using a [trio backend](https://anyio.readthedocs.io/en/stable/basics.html#running-async-programs).
From what I can see, pyzmq only [supports asyncio](https://github.com/zeromq/pyzmq/blob/5653496aa220cc39f414fbbc99962620fa63186b/zmq/asyncio.py). How do you feel about supporting AnyIO?

### Code to reproduce bug

_No response_

### Traceback, if applicable

_No response_

### More info

_No response_

davidbrochart:
Closes #1827.

### DIFF
--- a/tests/test_asyncio.py
+++ b/tests/test_asyncio.py
@@ -11,10 +11,18 @@ from concurrent.futures import CancelledError
 from multiprocessing import Process
 
 import pytest
+from anyio import create_task_group, move_on_after, sleep
+from anyioutils import CancelledError, create_task
 from pytest import mark
 
 import zmq
 import zmq.asyncio as zaio
+
+if sys.version_info < (3, 11):
+    from exceptiongroup import BaseExceptionGroup, ExceptionGroup
+
+
+pytestmark = pytest.mark.anyio
 
 
 @pytest.fixture
@@ -46,23 +54,17 @@ def test_instance_subclass_second(context):
 async def test_recv_multipart(context, create_bound_pair):
     a, b = create_bound_pair(zmq.PUSH, zmq.PULL)
     f = b.recv_multipart()
-    assert not f.done()
     await a.send(b"hi")
-    recvd = await f
-    assert recvd == [b"hi"]
+    assert await f == [b"hi"]
 
 
 async def test_recv(create_bound_pair):
     a, b = create_bound_pair(zmq.PUSH, zmq.PULL)
     f1 = b.recv()
     f2 = b.recv()
-    assert not f1.done()
-    assert not f2.done()
     await a.send_multipart([b"hi", b"there"])
-    recvd = await f2
-    assert f1.done()
-    assert f1.result() == b"hi"
-    assert recvd == b"there"
+    assert await f1 == b"hi"
+    assert await f2 == b"there"
 
 
 @mark.skipif(not hasattr(zmq, "RCVTIMEO"), reason="requires RCVTIMEO")
@@ -72,11 +74,11 @@ async def test_recv_timeout(push_pull):
     f1 = b.recv()
     b.rcvtimeo = 1000
     f2 = b.recv_multipart()
-    with pytest.raises(zmq.Again):
+    with pytest.raises(ExceptionGroup) as excinfo:
         await f1
+    assert excinfo.group_contains(zmq.Again)
     await a.send_multipart([b"hi", b"there"])
     recvd = await f2
-    assert f2.done()
     assert recvd == [b"hi", b"there"]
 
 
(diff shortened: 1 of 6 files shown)


minrk:
Awesome, thanks! Let me know when you think it's ready for review or if you have any questions. The edge-triggered FD stuff is very fiddly.

davidbrochart:
It's still a WIP, but it would be nice to know what you think about it @minrk.
The API in `sugar/socket.py` should probably not be async, I'm not sure exactly where to have the sync/async split.

minrk:
Yeah, `sugar/socket.py`, which is the base blocking API, should probably not need any changes at all, and definitely shouldn't introduce any async code.

It may be that the coroutine/wait_readable approach is different enough in design that it would be better to start `zmq/anyio.py` from scratch and only take _inspiration_ from the current asyncio implementation (or copy/edit), rather than updating in-place. That would also be the easiest way to ensure backward compatibility, which is important. But if it seems easier to update in-place, that's a possibility, too. It's hard for me to say since I don't know what it's going to look like.

The hardest part, I think, will be correctly handling the edge-triggered FD with multiple awaiters, especially since it is hard to ensure we provoke this situation thoroughly in tests. Any event that consumes socket events must necessarily wake all tasks that are waiting on the FD or they can hang. We'll need to be thorough about the implications of a socket being used in multiple task groups, because _all_ waiting tasks must wake. If that's a problem, we'll need to make sure to have locks to prevent the situations that won't work from happening.

Needing to wake tasks includes accessing the `socket.EVENTS` attribute, sending a message, and receiving a message. The FD being readable does _not_ mean the socket is readable, it means there _have been_ events on the socket, and `socket.EVENTS` must be checked. The socket not being readable _does not_ mean there are not messages waiting, it means that events have not changed. so e.g. with multiple messages ready to recv, the FD will wake once, and all messages must be received before it is appropriate to call `wait_readable` again.

Plus we'll need a `wait_readable` implementation that works on proactor on Windows before it can be used. We have the `add_reader` version of this via tornado, which will need to be adapted to `wait_readable`. The principle is the same, so it can be built on top, I think.

tests/test_asyncio.py
```diff
@@ -72,82 +74,70 @@ async def test_recv_timeout(push_pull):
...
-    assert f2.done()
     assert recvd == [b"hi", b"there"]
 
 
 @mark.skipif(not hasattr(zmq, "SNDTIMEO"), reason="requires SNDTIMEO")
 async def test_send_timeout(socket):
     s = socket(zmq.PUSH)
     s.sndtimeo = 100
-    with pytest.raises(zmq.Again):
+    with pytest.raises(ExceptionGroup) as excinfo:
```
minrk, line 89: Hopefully this isn't required when it's wrapped up. A single coroutine should raise a single error, right? This would be a major breaking change and a significant degradation of the API.

minrk:
Here's a [very simple anyio example](https://gist.github.com/minrk/cd0db83f99106afa875d5b6c87abf156) implementing async recv using only the blocking pyzmq API. It doesn't handle the edge-triggered nuances or timeouts, but it gets the very basics.

zmq/_future.py
```diff
@@ -260,24 +241,25 @@ def close(self, linger: int | None = None) -> None:
 
     def get(self, key):
         result = super().get(key)
-        if key == EVENTS:
-            self._schedule_remaining_events(result)
+        # if key == EVENTS:
+        #     self._schedule_remaining_events(result)
```
minrk, line 245: This is one of the things we will need to keep some version of. This is part of handling the edge-triggered FD: all tasks waiting on `wait_readable` must wake if EVENTS is accessed from anywhere.

davidbrochart:
Yes I think you're right that it's safer to have a separate implementation just for AnyIO, for backwards-compatibility reasons. I'm not sure anything I've done will remain then, since I was hoping to replace the asyncio layer with AnyIO, but now we can start from scratch as you did in your example.
This also raises the question of whether to create a different package, or have the AnyIO support here in pyzmq?

minrk:
I think either way is fine for the separate package or here. A separate package can also make sense as a prototyping pathway to release often and avoid needing to wait for pyzmq releases. It really _shouldn't_ need changes in pyzmq, though I do occasionally have to do some cooperation to get all the serialization methods to work well with the async subclasses without too much duplication.

For example, outside pyzmq it would be totally appropriate to _only_ support low-level send/recv_multipart, not all the serialization extensions. That doesn't fit in the pyzmq socket subclass approach where all methods need to work, and use a "has a" rather than "is a" relationship, which type systems like a lot better since subclass != subtype.

 We can always merge upstream here if/when it seems like 'the way to go' for async pyzmq, or not. We did that with `gevent_zeromq` and `pyzmq_static` before. Or if we like maintaining the separate package, gently deprecate `zmq.asyncio` in favor of the new package.

minrk:
(btw, what you have can still remain, and you can _copy_ the changed files to `zmq/anyio.py` and then feel free to update without concerns around breaking APIs). But if you'd rather start the whole design from scratch, that's also fine by me. Getting all the waits/timeouts/cleanup right was hard, but maybe (hopefully!) it will be easier with an anyio approach. Timeouts and cleanup are usually a lot easier with task groups and coroutines than futures! And if we had normal level-triggered FDs, this whole thing would be pretty straightforward.

Maybe it should even be a function-based approach (`async_send` / `async_recv`) using blocking Socket objects instead of defining async Socket classes (this might make the bookkeeping of waking multiple waiters hard, I'm not sure). I'm absolutely not set on using the _same_ approach I used for `zmq.asyncio` for anyio. I'm happy to pursue whatever you think would be suit anyio compatibility best.

davidbrochart:
One simple way would be to run pyzmq's blocking APIs in a thread, but I guess that would lead to very bad performances.

davidbrochart:
I have started https://github.com/davidbrochart/zmq-anyio so I'm closing this for now. Thanks for the feedback @minrk !

PR #2051, opened 2024-12-03
limited api wheels for 3.12+ with cython 3.1

### LINKED ISSUE #2010: adopt limited API for wheels
When Cython 3.1 is out, we can try using the limited API, which would let us publish many fewer wheels (108 for pyzmq 26.1).

My thought is to perhaps go with some version of SPEC0 for 'full' support:

- only build one limited API wheel for lowest supported Python with lowest supported glibc (currently 3.7, manylinux 2010, I think)
- don't build any wheels for other Pythons not supported by [SPEC0](https://scientific-python.org/specs/spec-0000/) - this would mean no more wheels needed for CPython 3.8-3.10
- possibly build a 3.12 limited wheel for performance reasons ([ref](https://docs.cython.org/en/latest/src/userguide/limited_api.html#building-with-the-limited-api))

This would reduce the CPython builds from 8 supported CPythons times ~12 targets to just 5 (2x 3.13).

If there are no downsides to limited wheels after 3.12, we could stop building wheels for each minor version altogether (hopefully free-threading will have a limited API again in 3.14).

If we went with _just_ limited API wheels, we'd only have to build for ~3 CPythons (3.7, 3.12, 3.13t).

minrk:
try building wheels with limited API

requires Cython 3.1, currently in prerelease

see #2010 for more about which wheels we could/should build

- I think we should definitely drop CPython 3.8-3.10, leaving those to the Stable ABI 3.7 wheel
- The _most_ we could drop is everything but 3.7 and 3.12 (and 3.13t)

closes #2010


minrk:
Actually _can't_ drop any wheels before 3.11 because 3.11 introduces Py_buffer to the ABI, which we need. So we still need to build wheels for each supported Python <=3.10.

So we can't drop most wheels, but we could drop CPython >= 3.13 or 3.12 in favor of a limited API wheel. It's unclear if the vectorcall improvement in 3.12 is worth it or not.

CMakeLists.txt
```diff
@@ -404,9 +404,19 @@ endif()
 
 file(MAKE_DIRECTORY ${ZMQ_BACKEND_DEST})
 
+if(NOT "${SKBUILD_SABI_COMPONENT}" STREQUAL "")
+  # set stable API
+  # assume we are targeting >= current Python version
+  # this isn't required, but we can't seem to get `wheel.py-api
+  message("SKBUILD_LIMITED_API=${SKBUILD_LIMITED_API}")
```
jcfr, line 411: It turns out that `SKBUILD_LIMITED_API` should have originally been documented as `SKBUILD_LIMITED_ABI`, then it was later removed in favor of checking if `SKBUILD_SABI_COMPONENT` is not empty. See https://github.com/scikit-build/scikit-build-core/pull/957/files#r1873869308

minrk:
cibuildwheel doesn't build wheels if there's already a compatible one, so if we want to build both a 3.11 abi3 wheel and a 3.13 native wheel, we need to do something special with ci (i.e. build abi3 wheels in their own jobs).

Running with Python 3.13, I saw a ~20% performance penalty of about 1µs per send (baseline 4.5µs) in the throughput benchmarks with the 3.11 abi (basically just measuring the overhead of `Socket.send`). the penalty was reduced to about 0.7µs for the 3.12 abi with vectorcall.

So I can either:

1. publish a stable abi wheel for 3.11, stop building newer wheels (~1µs penalty)
2. publish standard wheel for 3.11, stable for 3.12+ (~0.7µs penalty)
3. keep publishing all 'current' wheels, but publish _just_ the 3.12 stable abi, so that future Pythons can have a wheel before PyPI allows uploads (_more_ wheels, not fewer)

I'm not sure which I'll pick when the time comes, but perhaps option 2.

I _really_ hope freethreading gets a stable ABI in 3.14, too.

PR #2053, opened 2025-01-13
fix PYZMQ_NO_BUNDLE option name

mgorny:
Fix the CMake `option()` name from `PYZMQ_LIBZMQ_NO_BUNDLE` to `PYZMQ_NO_BUNDLE`.  The latter is actually used below as the environment variable and as the actual condition, as well as in the documentation. Given that there are projects already using it, just fix the name passed to `option()` to match.


minrk:
Thanks!

mgorny:
Thank you!

PR #2079, opened 2025-03-11
allow building cffi on cpython

### LINKED ISSUE #2076: FEAT: Please support building CFFI backend on CPython
Currently, pyzmq always use the Cython backend on CPython and CFFI backend on PyPy. While generally this works, it means that if we're seeing issues, it might hard to tell if they're related to PyPy itself or to CFFI (vs. Cython) extension.

Could you consider having an option to build the CFFI backend on CPython as well? The main use case would be being able to test the CFFI backend independently of PyPy.

Looking at `CMakeLists.txt`, I see two options:
- the easy option of adding an option like `PYZMQ_FORCE_CFFI` that would build only the CFFI backend,
- the hard option of adding an option like `PYZMQ_BUILD_CFFI` that would build both backends — it would require some more changes, such as moving the common bits to a helper function and calling them for both backends.

I'm ready to write a patch for either of these if you agree.

minrk:
with `PYZMQ_BACKEND=cffi`, the same longstanding env for selecting the backend at runtime:

```bash
PYZMQ_BACKEND=cffi pip wheel .
pip install *.whl cffi
```

Not everything works, notably `memoryview(Frame)` doesn't work because CPython doesn't support PyPy's `__buffer__()`. `cffi` build-time dependency injection works (thanks @mgorny for the tip!), but there doesn't appear to be one for runtime, so you have to install cffi, but that seems fine.

not really supported, but it should work. Does this work for you @mgorny?

closes #2076

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -201,6 +201,7 @@ jobs:
         if: matrix.backend
         run: |
           echo "PYZMQ_BACKEND=${{ matrix.backend }}" >> "$GITHUB_ENV"
+          pip install cffi
 
       - name: install libzmq-dev
         if: matrix.zmq == 'head'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -84,6 +84,10 @@ jobs:
 
           - os: ubuntu-24.04
             python: "3.13"
+            backend: cffi
+
+          - os: ubuntu-24.04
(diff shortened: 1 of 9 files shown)


mgorny:
Thanks. I think it should work. I'm sorry for not looking into it yet, I've gotten a pretty bad infection.

PR #2114, opened 2025-08-03
try ios wheels

minrk:


minrk:
These take a while and I've no idea if they are useful, so I may not merge this, but good to know they work

PR #2177, opened 2026-04-21
add missing `__all__` to type stubs

jorenham:

---

According to the [Python typing spec](https://typing.python.org/en/latest/guides/writing_stubs.html#all):

> A stub file should contain an `__all__` variable if and only if it is also present at runtime. In that case, the contents of `__all__` should be identical in the stub and at runtime. If the runtime dynamically adds or removes elements (for example if certain functions are only available on some system configurations), include all possible elements in the stubs.

I also verified that `stubtest` now doesn't complain about these `__all__` *anymore* (with `stubtest --mypy-config-file=mypy.ini zmq` to be precise).

This should also help tools like typestats to index the public API, which at the moment isn't all too accurate: https://jorenham.github.io/typestats/dashboard/report/#pyzmq


minrk:
Thanks for the PR! I'll have to think about how to maintain this going forward, since it seems very tedious.

jorenham:
> Thanks for the PR! I'll have to think about how to maintain this going forward, since it seems very tedious.

Maybe stubtest can help with that; because it's able to detect `__all__` mismatches.

minrk:
Yeah, I should probably be running that here

PR #2182, opened 2026-05-08
Fill in the missing type annotations in `zmq.utils.*`

jorenham:
This fills in the missing type annotations of the public symbols in `zmq.utils.*`, increaing the global [type coverage](https://jorenham.github.io/typestats/dashboard/report/#pyzmq) of `zmq` by 3.78% (as reported by `uvx typestats check pyzmq`)

Admittedly this isn't the most interesting of static typing improvements. But my goal is to submit other PRs as well to increase `zmq`'s type coverage even more :)

---

### DIFF
--- a/zmq/utils/garbage.py
+++ b/zmq/utils/garbage.py
@@ -5,29 +5,39 @@ used in zero-copy sends.
 # Copyright (C) PyZMQ Developers
 # Distributed under the terms of the Modified BSD License.
 
+from __future__ import annotations
+
 import atexit
 import struct
 import warnings
-from collections import namedtuple
 from os import getpid
 from threading import Event, Lock, Thread
+from typing import Final, NamedTuple
 
 import zmq
 
-gcref = namedtuple('gcref', ['obj', 'event'])
+
+class gcref(NamedTuple):
+    obj: object
+    event: Event | None
 
 
 class GarbageCollectorThread(Thread):
     """Thread in which garbage collection actually happens."""
 
-    def __init__(self, gc):
+    gc: GarbageCollector
+    daemon: bool
+    pid: int
+    ready: Event
+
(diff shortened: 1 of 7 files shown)


zmq/utils/garbage.py
```diff
@@ -185,7 +199,7 @@ def is_alive(self):
             return False
         return True
 
-    def store(self, obj, event=None):
+    def store(self, obj: object, event: Event | None = None) -> int:
```
minrk, line 202:
```suggestion
    def store(self, obj: Any, event: Event | None = None) -> int:
```

zmq/utils/jsonapi.py
```diff
@@ -17,15 +17,15 @@
 jsonmod = json
 
 
-def dumps(o: Any, **kwargs) -> bytes:
+def dumps(o: object, **kwargs: Any) -> bytes:
```
minrk, line 20:
```suggestion
def dumps(o: Any, **kwargs: Any) -> bytes:
```

zmq/utils/garbage.py
```diff
@@ -5,29 +5,39 @@
...
 from os import getpid
 from threading import Event, Lock, Thread
+from typing import Final, NamedTuple
 
 import zmq
 
-gcref = namedtuple('gcref', ['obj', 'event'])
+
+class gcref(NamedTuple):
+    obj: object
```
minrk, line 21: is there a reason this is `object` and not `Any`?

jorenham: yes; `Any` is intrinsically type-unsafe, so it's best to avoid it if possible. And in these cases, there is no difference between `Any` or `object`, making `object` therefore the better choice.

minrk: good to know, thanks!

jorenham:
Thanks for the review!

PR #2183, opened 2026-05-19
Fill in the missing type annotations in `asyncio`, `auth`, `constants`, `decorators`, `error`

jorenham:
Similar to #2182, this increases the type coverage of `zmq` by filling missing type annotations of public symbols in a subset if the `zmq` modules. With this, type coverage increases from 73.69% to 75.88% (+2.19%) (according to [typestats](https://github.com/jorenham/typestats)).

No runtime changes; this only affects (i.e. improves) static typing.

---

Thanks for submitting a PR!

LLM/AI contributions are not accepted.

Please make sure:

- [x] I wrote this (no AI-generated code or text is included in this PR or its description)
- [x] any code changes are tested, if possible

PR #2184, opened 2026-05-19
Typing fixes and improvements for `Socket`, `Frame`, and `Context`

jorenham:
This PR was initially intended for filling in the missing public type annotations in `zmq.backend`, like #2183 and #2182. But while doing so, I noticed several typing issues: incorrect return types, incompetible overrides, incorrect type variable variance, and some minor overload issues. Because, you see, mypy completely ignores unannotated functions by default. So by annotating these methods, several hidden typing issues surfaced. Those issues weren't as simple as fixing one thing; I had to fix a whole bunch of stuff to get everything nice and consistent.
This unfortunately also caused this to not be as self-contained as I had hoped it would be. I don't think there's a good way to split this up though, or at least, not in a way that mypy will like.

FWIW: I had Claude Opus 4.7 review (but not touch) every change here. After iterating like this for a bit, it now seems to fully agree with the changes. But don't worry, those reviews is *all* I used AI for :)

Besides the typing fixes, this also increases type coverage by 4.07%.

---

Thanks for submitting a PR!

LLM/AI contributions are not accepted.

Please make sure:

### DIFF
--- a/examples/mongodb/controller.py
+++ b/examples/mongodb/controller.py
@@ -71,7 +71,7 @@ class MongoZMQ:
             if len(msg) != 3:
                 error_msg = f'invalid message received: {msg}'
                 print(error_msg)
-                reply = [msg[0], error_msg]
+                reply = [msg[0], error_msg.encode()]
                 socket.send_multipart(reply)
                 continue
             id = msg[0]
@@ -81,11 +81,11 @@ class MongoZMQ:
             reply = [id]
             if operation == 'add':
                 self.add_document(contents)
-                reply.append("success")
+                reply.append(b"success")
             elif operation == 'get':
                 doc = self.get_document_by_keys(contents)
                 json_doc = self._doc_to_json(doc)
-                reply.append(json_doc)
+                reply.append(json_doc.encode())
             else:
                 print('unknown request')
             socket.send_multipart(reply)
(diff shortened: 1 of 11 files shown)


jorenham:
Hmm apparently I messed up somewhere... Marking this as draft until I fixed it

jorenham:
Well, as it turns out, these new mypy errors uncovered actual bugs in the examples. For example:

```bash
$ python examples/poll/reqrep.py 
Running polling tests for REQ/REP sockets...
  File "/home/joren/Workspace/pyzmq/examples/poll/reqrep.py", line 37, in <module>
    s2.send('msg1')
    ~~~~~~~^^^^^^^^
  File "/home/joren/Workspace/pyzmq/zmq/sugar/socket.py", line 713, in send
    return super().send(data, flags=flags, copy=copy, track=track)
           ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "zmq/backend/cython/_zmq.py", line 1152, in zmq.backend.cython._zmq.Socket.send
    def send(self, data, flags=0, copy: bint = True, track: bint = False):
  File "zmq/backend/cython/_zmq.py", line 1197, in zmq.backend.cython._zmq.Socket.send
    raise TypeError("unicode not allowed, use send_string")
TypeError: unicode not allowed, use send_string
```

I'll fix the examples as well then.

PR #2185, opened 2026-05-19
Fix `Socket.send` return type docstrings

jorenham:
While working on #2184 I noticed that the docstrings for the return type of the `send`  methods of the `Socket`s wasn't quite right. So to avoid #2184 from getting even bigger than it already is, I figured that this documentation fix could be addressed separately.

---

No AI, btw

### DIFF
--- a/zmq/backend/cython/_zmq.py
+++ b/zmq/backend/cython/_zmq.py
@@ -1174,11 +1174,11 @@ class Socket:
 
         Returns
         -------
-        None : if `copy` or not track
-            None if message was sent, raises an exception otherwise.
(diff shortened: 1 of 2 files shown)


minrk:
Thanks for noticing! I need to think about whether the docstring should be updated or the code should be updated to match the docstring. I think it might be a bug that no-copy sends are used in `send(Frame, copy=True)` and the docstring describes what _should_ happen. There might have been a performance issue along the way that copying Frames wasn't worth it. I'm not sure! There is a memory safety issue if message buffers aren't copied when requested.

PR #2190, opened 2026-05-19
Support for mypy 2.1.0

jorenham:
There are a couple of (related) things going on here:

- bump mypy in pre-commit and the test requirements (but keep using mypy ~1.20~ 1.19 on EOL Python 3.9)
- added typeshed typing stubs to both pre-comment and test dependencies, eliminating the differences between them
- configured mypy to warn about unused ignores, configs, and casts
- removed several unused ignores and casts

Locally everything seems to pass with both mypy ~1.20.2~ 1.19.1 and 2.1.0. Once Python 3.9 support is dropped so we don't have to worry about mypy<2 compat anymore, but at the moment this doesn't seem to be much of a problem (on my machine, that is).

---

There are still many `type: ignore`s left though, and there are quite a few that seem to be caused by bugs in mypy.

So if you want, I could open a PR showing what it'll look like if we'd switched from mypy to, say, Pyrefly (which is a good option at the moment in my experience). Type-checking will be quite a bit faster with Pyrefly, and it's also very handy that Pyrefly comes with an LSP, so it's a lot nicer to work with in <insert your IDE here> (including neovim and emacs).

Using both pyrefly and mypy is also an option; they usually agree with each other (although Pyrefly is currently a bit more "correct" according to the Python type spec conformance suite). In NumPy, for example, we do exactly this: use both mypy and pyrefly.

Anyway, let me know if I should open a Pyrefly PR to see what it'll look like.

---

Thanks for submitting a PR!

LLM/AI contributions are not accepted.

Please make sure:


minrk:
Thanks for updating this!

I don't have any particular interest in mypy vs pyrefly, but personally I'd rather stick with the community-maintained apparent standard. If Facebook's pyrefly takes over as the de facto standard, then migrating seems okay, I suppose. I definitely don't want to deal with more than one type checker.

